PR #3, opened 2021-03-18
Initialize subprojects and command-line interface

### LINKED ISSUE #2: Prepare a native command-line interface for save framework
Now it can have only `--help` option

petertrr:
### What's done:
* Added subprojects
* Added CLI runner
* Added core classes
* Added versions and talaiot gradle plugins

Closes #2


save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlinx.cli.default
+
+/**
+ * @param args CLI args
+ * @return an instance of [SaveConfig]
+ */
+fun createConfigFromArgs(args: Array<String>): SaveConfig {
+    val parser = ArgParser("save")
+
+    val configPath by parser.option(
```
akuleshov7, line 23: just "config", cli should not have complex argument names. Also camel case is not applicable for cli

save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
```diff
@@ -0,0 +1,53 @@
...
+        ArgType.String,
+        shortName = "f",
```
akuleshov7, line 25: -c

save-core/src/commonMain/kotlin/org/cqfn/save/core/config/ReportType.kt
```diff
@@ -0,0 +1,8 @@
+package org.cqfn.save.core.config
+
+/**
+ * Possible types of output formats.
+ */
+enum class ReportType {
+    JSON, XML
```
akuleshov7, line 7: TOML 
PLAIN

save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import kotlinx.cli.ArgParser
+import kotlinx.cli.ArgType
+import kotlinx.cli.default
+
+/**
+ * @param args CLI args
+ * @return an instance of [SaveConfig]
+ */
+fun createConfigFromArgs(args: Array<String>): SaveConfig {
```
akuleshov7, line 20: may be will add some argument for a baseline? (while I haven't yet forgotten about that)
We should have a baseline functionality

akuleshov7 (changes requested):
please see comments and we will merge it as a basic template concept

save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
petertrr: Added this and addressed other comments, please take a look

akuleshov7 (approved):
lgtm

PR #19, opened 2021-04-05
Detect save.toml config hierarchy

### LINKED ISSUE #13: Support detecting configs and tests in the test repository
We should support the detection of save.toml configuration and support the hierarchy of them (save.toml configs should be inherited from the higher levels of directories)

petertrr:
### What's done:
* Added parent and child configs to TestSuiteConfig
* Added logic in ConfigDetector.kt
* Added FileUtils.kt for common operations on files using okio
* Added tests
* Compile for Java 11

Closes #13

### DIFF
--- a/save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
+++ b/save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
@@ -36,4 +36,4 @@ data class TestSuiteConfig(
 /**
  * @return whether a file denoted by this [Path] is a default save configuration file (save.toml)
  */
-fun Path.isDefaultSaveConfig() = name == "save.toml"
+fun Path.isSaveTomlConfig() = name == "save.toml"
--- a/save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
+++ b/save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
@@ -8,12 +8,14 @@ import okio.Path
  * @property description description of test suite
  * @property location [Path] denoting the location of this file
  * @property parentConfig parent config in the hierarchy of configs, `null` if this config is root.
+ * @property childConfigs list of child configs in the hierarchy of configs, can be empty if this config is at the very bottom.
  */
 data class TestSuiteConfig(
     val suiteName: String,
     val description: String,
     val location: Path,
-    val parentConfig: TestSuiteConfig?
+    val parentConfig: TestSuiteConfig?,
+    val childConfigs: MutableList<TestSuiteConfig> = mutableListOf(),
 ) {
     /**
      * @return whether this config file is in the root on the hierarchy
--- a/save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
+++ b/save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
@@ -8,24 +8,29 @@ import okio.Path
  * @property description description of test suite
  * @property location [Path] denoting the location of this file
  * @property parentConfig parent config in the hierarchy of configs, `null` if this config is root.
- * @property childConfigs list of child configs in the hierarchy of configs, can be empty if this config is at the very bottom.
  */
 data class TestSuiteConfig(
     val suiteName: String,
     val description: String,
     val location: Path,
     val parentConfig: TestSuiteConfig?,
-    val childConfigs: MutableList<TestSuiteConfig> = mutableListOf(),
 ) {
+    /**
+     * List of child configs in the hierarchy of configs, can be empty if this config is at the very bottom.
+     * NB: don't move to constructor in order not to break toString into infinite recursion.
+     */
+    val childConfigs: MutableList<TestSuiteConfig> = mutableListOf()
+
     /**
      * @return whether this config file is in the root on the hierarchy
      */
     fun isRoot() = parentConfig == null
 
     /**
+     * @param wihSelf if true, include this config as the first element of the sequence or start with parent config otherwise
      * @return a [Sequence] of parent config files
      */
-    fun parentConfigs() = generateSequence(parentConfig) { it.parentConfig }
+    fun parentConfigs(wihSelf: Boolean = false) = generateSequence(if (wihSelf) this else parentConfig) { it.parentConfig }
 }
 
 /**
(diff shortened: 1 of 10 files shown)


save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
```diff
@@ -1,11 +1,39 @@
...
+    /**
+     * @return whether this config file is in the root on the hierarchy
+     */
+    fun isRoot() = parentConfig == null
+
+    /**
+     * @param wihSelf if true, include this config as the first element of the sequence or start with parent config otherwise
+     * @return a [Sequence] of parent config files
+     */
+    fun parentConfigs(wihSelf: Boolean = false) = generateSequence(if (wihSelf) this else parentConfig) { it.parentConfig }
```
akuleshov7, line 33: I thought that we will store here all parental configs, but yes, this looks much better

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.cqfn.save.core.config.TestSuiteConfig
+import org.cqfn.save.core.config.isDefaultSaveConfig
+
+import okio.FileSystem
+import okio.Path
+
+/**
+ * A class that is capable of discovering config files hierarchy.
+ */
+class ConfigDetector {
```
akuleshov7, line 12: why do we need class here... ?

petertrr: In case we decide to enhance this logic in the future, mainly. This logic could be added to the `TestSuiteConfig` itself, but I decided it would look better in a separate class (and look weird as a bunch of top-level functions)

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
```diff
@@ -0,0 +1,79 @@
...
+                                        locationsFlattened.take(index + 1).reversed().find { it.parent in path.parents() }!!
+                            }!!
+                        ).also {
+                            it.parentConfig!!.childConfigs.add(it)
+                        }
+                    )
+                }
+        }
+
+    private fun discoverConfigWithParents(file: Path): TestSuiteConfig? = when {
```
akuleshov7, line 48: do we need to add logging in case of no detection?

akuleshov7 (changes requested):
General comment: please add proper logging. 
I have discovered this. I have discovered that, e.t.c

save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestSuiteConfig.kt
```diff
@@ -1,11 +1,39 @@
...
+}
+
+/**
+ * @return whether a file denoted by this [Path] is a default save configuration file (save.toml)
+ */
+fun Path.isDefaultSaveConfig() = name == "save.toml"
```
akuleshov7, line 39: I don't think that this (the name) really will be changed one day.

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
```diff
@@ -0,0 +1,79 @@
...
+            val configs = mutableListOf(config)
+            locationsFlattened
+                .drop(1)  // because [config] will be discovered too
+                .forEachIndexed { index, path ->
+                    configs.add(
+                        TestSuiteConfig(
+                            "todo: read from file",
+                            "todo: read from file",
+                            path,
+                            configs.find { discoveredConfig ->
```
akuleshov7, line 37: pls calculate it outside of constructor

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
```diff
@@ -0,0 +1,79 @@
...
+                                discoveredConfig.location ==
+                                        locationsFlattened.take(index + 1).reversed().find { it.parent in path.parents() }!!
+                            }!!
+                        ).also {
+                            it.parentConfig!!.childConfigs.add(it)
```
akuleshov7, line 42: parent config can be null here, I guess

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
```diff
@@ -0,0 +1,79 @@
...
+        else -> null
+    }
+
+    private fun testSuiteConfigFromFile(file: Path): TestSuiteConfig = TestSuiteConfig(
+        "todo: read from file",
+        "todo: read from file",
+        file,
+        file.parents()
+            .drop(1)  // because immediate parent already contains [this] config
+            .mapNotNull { parentDir ->
```
akuleshov7, line 71: so you do mapNotNull, but in this map can return null in findChildByNull

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
petertrr: No, because it's a loop among descendants of current config, so each of them will have a parent

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
petertrr: Yes, this way we can skip parent directories that don't contain a config file and find it on higher levels

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/ConfigDetector.kt
akuleshov7: as you wish :)

save-core/src/commonMain/kotlin/org/cqfn/save/core/logging/Logger.kt
```diff
@@ -0,0 +1,36 @@
...
+ * @param msg a message string
+ */
+fun logInfo(msg: String): Unit = println("INFO: $msg")
+
+/**
+ * Log a message with error level
+ *
+ * @param msg a message string
+ */
+fun logError(msg: String): Unit = println("ERROR: $msg")
```
akuleshov7, line 36: where are warnings?

akuleshov7 (approved):
lgtm

save-core/src/commonMain/kotlin/org/cqfn/save/core/logging/Logger.kt
petertrr: Added

PR #23, opened 2021-04-09
Detect save.toml config hierarchy from properties

### LINKED ISSUE #15: Support detection of tests in the repository by save.properties file
Save should find, detect and configure tests using `save.properties` file

petertrr:
### What's done:
* Added logic
* Fixed README

Closes #15

### DIFF
--- a/save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
+++ b/save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
@@ -10,17 +10,19 @@ import org.cqfn.save.core.config.ResultOutputType
 import org.cqfn.save.core.config.SaveConfig
 import org.cqfn.save.core.logging.logDebug
 import org.cqfn.save.core.logging.logError
+import org.cqfn.save.core.logging.logInfo
 
 import okio.FileSystem
 import okio.IOException
 import okio.Path.Companion.toPath
 
+import kotlinx.cli.AbstractSingleOption
 import kotlinx.cli.ArgParser
 import kotlinx.cli.ArgType
+import kotlinx.cli.MultipleOption
 import kotlinx.cli.default
 import kotlinx.cli.multiple
 import kotlinx.cli.required
-import org.cqfn.save.core.logging.logInfo
 
 private fun <U> Map<String, String>.getAndParseOrElse(
     key: String,
@@ -36,14 +38,16 @@ private fun <U> Map<String, String>.getAndParseOrElse(
 @Suppress("TOO_LONG_FUNCTION")
 fun createConfigFromArgs(args: Array<String>): SaveConfig {
     // The first iteration of parsing - get location of properties file. Ignore any other arguments.
-    val propertiesFile = args.toList().zipWithNext()
-        .firstOrNull { it.first == "--properties-file" || it.first == "-prop" }
+    val propertiesFileOptionFullName = "properties-file"
+    val propertiesFileOptionShortName = "prop"
+    val propertiesFileName = args.toList().zipWithNext()
+        .firstOrNull { it.first == "--$propertiesFileOptionFullName" || it.first == "-$propertiesFileOptionShortName" }
         ?.second
         ?: "save.properties"
-    logDebug("Using properties file $propertiesFile")
+    logDebug("Using properties file $propertiesFileName")
 
     val properties: Map<String, String> = try {
-        FileSystem.SYSTEM.read(propertiesFile.toPath()) {
+        FileSystem.SYSTEM.read(propertiesFileName.toPath()) {
             generateSequence { readUtf8Line() }.toList()
         }
             .associate { line ->
@@ -52,8 +56,8 @@ fun createConfigFromArgs(args: Array<String>): SaveConfig {
                 }
             }
     } catch (e: IOException) {
-        logError("Unable to read properties file $propertiesFile: ${e.message}")
-        throw e  // todo exit with exit code
+        logError("Unable to read properties file $propertiesFileName: ${e.message}")
+        emptyMap()  // todo exit with exit code?
     }
     logInfo("Read from properties file: $properties")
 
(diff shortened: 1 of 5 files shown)


save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
```diff
@@ -8,72 +8,103 @@ import org.cqfn.save.core.config.LanguageType
...
+        logError("Unable to read properties file $propertiesFile: ${e.message}")
+        throw e  // todo exit with exit code
+    }
+    logInfo("Read from properties file: $properties")
+
+    val parser = ArgParser("save")  // todo: or save-cli?
 
     val config by parser.option(
         ArgType.String,
+        fullName = "config",
```
kgevorkyan, line 68: Why did you add fullName in this case? Is it redundant, as we already use same name of the `val config`?

petertrr: Yes, you're right. Removed.

save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
```diff
@@ -8,79 +8,124 @@ import org.cqfn.save.core.config.LanguageType
...
+            generateSequence { readUtf8Line() }.toList()
+        }
+            .associate { line ->
+                line.split("=", limit = 2).let {
+                    it.first() to it.last()
+                }
+            }
+    } catch (e: IOException) {
+        logError("Unable to read properties file $propertiesFileName: ${e.message}")
+        emptyMap()  // todo exit with exit code?
```
akuleshov7, line 60: yep, I think that you just should be doing `exitProcess`

kgevorkyan (approved):

save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
```diff
@@ -8,79 +8,124 @@ import org.cqfn.save.core.config.LanguageType
...
+    logInfo("Read from properties file: $properties")
+
+    val parser = ArgParser("save")  // todo: or save-cli?
 
     val config by parser.option(
         ArgType.String,
         shortName = "c",
         description = "Path to the root save config file",
-    ).default("save.toml")
+    ).default(properties.getOrElse("config") { "save.toml" })
```
akuleshov7, line 70: no-no, let's make a good deserialization to the class here instead of "getOrElse" and string

petertrr: Some options can be passed via CLI, some others - via the file, so there is no way to deserialize all fields at once (without writing a complicated deserializer). Or we would need a mutable counterpart of SaveConfig, which will be partially deserialized from file and then filled with CLI options.

akuleshov7:
```
class MapDecoder(val map: Map<*, *>, var elementsCount: Int = 0) : AbstractDecoder() {
        private var elementIndex = 0

        override val serializersModule: SerializersModule = EmptySerializersModule

        override fun decodeValue(): Any {
            // of course you should raise a good error here if in your map with values from input you won't find
            // enough values for your serializable class. For example if you class has more fields than there are in the input
            // and these fields are not optional - you should raise an error
            val keyAtTheCurrentIndex = map.keys.elementAt(elementIndex - 1)
            return map[keyAtTheCurrentIndex]!!
        }

        /**
         * this method should be overriden to map between the FIELDS in your class and the VALUE from
         * the input that you would like to inject into it
         */
        override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
            if (elementIndex == map.size) return CompositeDecoder.DECODE_DONE

            // index of the field from the class where we should inject our value
            val fieldWhereValueShouldBeInjected =
                descriptor.getElementIndex(map.keys.elementAt(elementIndex).toString())

            if (fieldWhereValueShouldBeInjected == CompositeDecoder.UNKNOWN_NAME) {
                // if (weAreStrict) {throw Exception("unknown property") }
            }
            elementIndex++

            return fieldWhereValueShouldBeInjected
        }

        override fun decodeCollectionSize(descriptor: SerialDescriptor): Int =
            decodeInt().also {
                elementsCount = it
            }

        // used to trigger the processing for structures (including nested)
        override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
            // corner case at the beginning of the decoding
            if (elementIndex == 0) {
                validateDecodedInput(descriptor, map)
                return MapDecoder(map, descriptor.elementsCount)
            } else {
                // need to decrement element index, as unfortunately it was incremented in the iteration of `decodeElementIndex`
                return when (val innerMap = map.values.elementAt(elementIndex - 1)) {
                    is Map<*, *> -> {
                        validateDecodedInput(descriptor, innerMap)
                        MapDecoder(innerMap, descriptor.elementsCount)
                    }
                    else -> throw Exception("Incorrect nested data")
                }
            }
        }

        override fun decodeNotNullMark(): Boolean = decodeString() != "NULL"

        // we should validate that all keys (required fields) from the class are not missing in the input
        private fun validateDecodedInput(descriptor: SerialDescriptor, map: Map<*, *>) {
            val missingKeysInInput = descriptor.elementNames.toSet() - map.keys.toSet()
            missingKeysInInput.forEach {
                val elementIndex = descriptor.getElementIndex(it.toString())
                if (!descriptor.isElementOptional(elementIndex)) {
                    throw Exception("Invalid number of arguments for deserialization, missing required field" +
                            " ${descriptor.getElementName(elementIndex)}")
                }
            }
        }
    }

    fun <T> decodeFromString(resMap: Map<*, *>, deserializer: DeserializationStrategy<T>): T {
        val decoder = MapDecoder(resMap)
        return decoder.decodeSerializableValue(deserializer)
    }
```

petertrr: Yep, *easy enough* 😨

akuleshov7 (approved):
lgtm

save-cli/src/commonMain/kotlin/org/cqfn/save/cli/Config.kt
petertrr: I'll do it in a separate PR, because code should be moved from commonMain to nativeMain

PR #26, opened 2021-04-09
Use exitProcess in save-cli

petertrr:
### What's done:
* Moved commonMain to nativeMain
* Added ExitCodes


save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/Config.kt
```diff
@@ -57,7 +58,7 @@ fun createConfigFromArgs(args: Array<String>): SaveConfig {
             }
     } catch (e: IOException) {
         logError("Unable to read properties file $propertiesFileName: ${e.message}")
-        emptyMap()  // todo exit with exit code?
+        exitProcess(ExitCodes.GENERAL_ERROR.code)
```
akuleshov7, line 61: I think this will not be the only place, where we should have this.
I guess that we need to unify it with logError into one function

petertrr: Added

akuleshov7 (approved):
lgtm

PR #27, opened 2021-04-13
diff-plugin

### LINKED ISSUE #18: Support the logic for the standard diff-plugin

petertrr:
### What's done:
* Introduced PluginConfig interface
* Added FixPlugin and FixPluginConfig
* Update usage details in README.md
* Added ProcessBuilder
* Added tests

Closes #18 

Known issues:
* There are two almost identical implementations of ProcessBuilder, because `popen` has slightly different signature on Linux and mingw


save-core/src/linuxX64Main/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -0,0 +1,39 @@
+@file:Suppress("FILE_WILDCARD_IMPORTS")
+
+package org.cqfn.save.core.utils
+
+import okio.FileSystem
+import okio.Path
+import platform.posix.*
```
kgevorkyan, line 7: Do we really need import *?

save-core/src/mingwX64Main/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -0,0 +1,39 @@
...
+    "MISSING_KDOC_CLASS_ELEMENTS",
+    "MISSING_KDOC_ON_FUNCTION"
+)
+actual class ProcessBuilder {
+    actual fun exec(command: List<String>, redirectTo: Path?): ExecutionResult {
+        val pd = popen!!.invoke(command.joinToString(" ").cstr.placeTo(MemScope()), "r".cstr.placeTo(MemScope()))
+
+        val stdout = buildString {
+            val buffer = ByteArray(4096)
+            while (true) {
```
kgevorkyan, line 21: We can't write 
`while (fgets(buffer, buffer.size, pd)` instead of `while (true)`? 
buffer doesn't collect output information and fgets doesn't return exit code?

save-core/src/linuxX64Main/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: Same situation in multiple places

petertrr: They use wildcard imports for posix and cinterop in official guides, but I agree, we prefer explicit imports usually. I'll change.

save-core/src/mingwX64Main/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
petertrr: Yep, I've forgotten how those C methods work :) Fixed.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    override fun execute(saveConfig: SaveConfig, testSuiteConfig: TestSuiteConfig) {
+        val fixPluginConfig = testSuiteConfig.pluginConfigs.filterIsInstance<FixPluginConfig>().single()
+        discoverFilePairs(fixPluginConfig.testResources)
+            .also {
+                logInfo("Discovered the following file pairs: $it")
+            }
+            .forEach { (expected, test) ->
+                pb.exec(fixPluginConfig.execCmd.split(" "), null)
+                val fixedLines = FileSystem.SYSTEM.readLines(
```
kgevorkyan, line 30: In my case IDE recommends to add annotation ExperimentalFileSystem

petertrr: I've added opt-in as a compiler flag, it should propagate this opt-in for the whole module. I believe you need to reimport the project.

kgevorkyan (approved):
lgtm

PR #33, opened 2021-04-16
Enable diktat for code in buildSrc

petertrr:
### What's done:
* Changed diktat config
* Fixed code style


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DiktatConfiguration.kt
```diff
@@ -17,14 +17,26 @@ fun Project.configureDiktat() {
     apply<DiktatGradlePlugin>()
     configure<DiktatExtension> {
         diktatConfigFile = rootProject.file("diktat-analysis.yml")
-        inputs = files("src/**/*.kt")
+        inputs = files("src/**/*.kt"
+//            , "**/*.kts"
```
kgevorkyan, line 21: Can we apply this pattern or may be we should delete this line, since it's not in use?

buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DiktatConfiguration.kt
```diff
@@ -17,14 +17,26 @@ fun Project.configureDiktat() {
...
  * Creates unified tasks to run diktat on all projects
  */
 fun Project.createDiktatTask() {
+    if (this == rootProject) {
+        // apply diktat to buildSrc
+        apply<DiktatGradlePlugin>()
+        configure<DiktatExtension> {
+            diktatConfigFile = rootProject.file("diktat-analysis.yml")
+            inputs = files("$rootDir/buildSrc/src/**/*.kt"
+            //    , "$rootDir/*.kts", "$rootDir/buildSrc/**/*.kts"
```
kgevorkyan, line 36: Same question

buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DiktatConfiguration.kt
petertrr: There are some false-positives, I've opened https://github.com/cqfn/diKTat/issues/834 for them. I'd say keep these as a todo for now, until it's fixed in diktat

kgevorkyan (approved):
lgtm

PR #35, opened 2021-04-16
Configuration for releases

### LINKED ISSUE #32: Create release process for the save framework

Cheshiriks:
What's done:
* Add Publishing Configuration files

Closes #32

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,68 @@
+name: Create release
+
+on:
+  push:
+    tags:
+      - 'v*'
+env:
+  GPG_SEC: ${{ secrets.PGP_SEC }}
+  GPG_PASSWORD: ${{ secrets.PGP_PASSWORD }}
+  OSSRH_USERNAME: ${{ secrets.SONATYPE_USER }}
+  OSSRH_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+
+jobs:
+  release_linux:
+    name: Build release on main platform (Linux)
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.11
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/PublishingConfiguration.kt
```diff
@@ -0,0 +1,147 @@
...
+        logger.lifecycle("The following publications are getting signed: ${extensions.getByType<PublishingExtension>().publications.map { it.name }}")
+        sign(*extensions.getByType<PublishingExtension>().publications.toTypedArray())
+    }
+}
+
+private fun Project.configureNexusPublishing() {
+    configure<NexusPublishExtension> {
+        repositories {
+            sonatype {
+                // only for users registered in Sonatype after 24 Feb 2021
```
petertrr, line 139: Need to check a url for older users, because cqfn has been registered at least a year ago

petertrr (changes requested):
Looks OK, but we'll need to test it somehow before merging. Also, I believe build will currently fail on macos because of ProcessBuilder

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: false
+          prerelease: false
+  release_macos:
+    name: Build release on MacOS
```
petertrr, line 40: We should disable this job until we support macos

petertrr:
Also you should remove `version` from `gradle.properties` if it is being set dynamically by reckon plugin

buildSrc/src/main/kotlin/org/cqfn/save/buildutils/PublishingConfiguration.kt
petertrr, line 139: Well, you can remove this comment now

petertrr (approved):

PR #36, opened 2021-04-20
Provide more detailed logging

kgevorkyan:
Provide more detailed logging
### What's done:
*  Add more log output over the project
*  Improve logic in ProcessBuilder: Move common part of code to the one class. Separate errors from popen and from executing command, add more logs

### DIFF
--- a/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
+++ b/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
@@ -35,9 +35,8 @@ fun SaveProperties.validate(): SaveProperties {
 /**
  * @return string which represents all fields of current instance
  */
-fun SaveProperties.getFields(): String {
-    return this.toString().dropWhile { it != '(' }.drop(1).dropLast(1)
-}
+fun SaveProperties.getFields() = this.toString().dropWhile { it != '(' }.drop(1)
+    .dropLast(1)
 
 /**
  * @param args CLI args
--- a/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
+++ b/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
@@ -33,6 +33,13 @@ fun SaveProperties.validate(): SaveProperties {
 }
 
 /**
+ * @return string which represents all fields of current instance
+ */
+fun SaveProperties.getFields(): String {
+    return this.toString().dropWhile { it != '(' }.drop(1).dropLast(1)
+}
+
+/**
  * @param args CLI args
  * @return an instance of [SaveProperties]
  */
@@ -40,10 +47,13 @@ fun SaveProperties.validate(): SaveProperties {
 fun createConfigFromArgs(args: Array<String>): SaveProperties {
     // getting configuration from command-line arguments
     val configFromCli = SaveProperties(args)
+    logDebug("Properties after parsed command line args:\n${configFromCli.getFields()}")
     // reading configuration from the properties file
     val configFromPropertiesFile = readPropertiesFile(configFromCli.propertiesFile)
-    // merging two configurations into signle [SaveProperties] class with a priority to command line arguments
-    return configFromCli.mergeConfigWithPriorityToThis(configFromPropertiesFile).validate()
+    // merging two configurations into single [SaveProperties] class with a priority to command line arguments
+    val mergedProperties = configFromCli.mergeConfigWithPriorityToThis(configFromPropertiesFile)
+    logInfo("Summary properties:\n${mergedProperties.getFields()}")
+    return mergedProperties.validate()
 }
 
 /**
@@ -54,7 +64,7 @@ fun createConfigFromArgs(args: Array<String>): SaveProperties {
 fun readPropertiesFile(propertiesFileName: String?): SaveProperties {
     propertiesFileName ?: return SaveProperties()
 
-    logDebug("Reading properties file $propertiesFileName")
+    logDebug("Reading properties from the file: $propertiesFileName")
 
     val properties: Map<String, String> = try {
         FileSystem.SYSTEM.read(propertiesFileName.toPath()) {
@@ -77,9 +87,6 @@ fun readPropertiesFile(propertiesFileName: String?): SaveProperties {
         )
     }
 
-    logDebug("Read properties from the properties file: $properties")
-    val deserializedPropertiesFile: SaveProperties = Properties.decodeFromStringMap(serializer(), properties)
-    logInfo("Read properties from the properties file: $deserializedPropertiesFile")
-
-    return deserializedPropertiesFile
+    logDebug("Found properties: $properties")
+    return Properties.decodeFromStringMap(serializer(), properties)
 }
--- a/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
+++ b/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
@@ -51,7 +51,7 @@ fun createConfigFromArgs(args: Array<String>): SaveProperties {
     val configFromPropertiesFile = readPropertiesFile(configFromCli.propertiesFile)
     // merging two configurations into single [SaveProperties] class with a priority to command line arguments
     val mergedProperties = configFromCli.mergeConfigWithPriorityToThis(configFromPropertiesFile)
-    logInfo("Summary properties:\n${mergedProperties.getFields()}")
+    logInfo("Using the following properties for SAVE execution:\n${mergedProperties.getFields()}")
     return mergedProperties.validate()
 }
 
--- a/save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
+++ b/save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
@@ -25,9 +25,9 @@ class Save(
         val plugins: List<Plugin> = emptyList()  // todo: discover plugins (from configuration blocks in TestSuiteConfig?)
         logInfo("Discovered plugins: $plugins")
         plugins.forEach {
-            logInfo("Execute plugin: ${it::class::simpleName}")
+            logInfo("Execute plugin: ${it::class.simpleName}")
             it.execute(saveProperties, testConfig)
-            logInfo("${it::class::simpleName} successfully executed!")
+            logInfo("${it::class.simpleName} successfully executed!")
         }
     }
 }
--- a/save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
+++ b/save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
@@ -2,6 +2,7 @@ package org.cqfn.save.core
 
 import org.cqfn.save.core.config.SaveProperties
 import org.cqfn.save.core.files.ConfigDetector
+import org.cqfn.save.core.logging.logInfo
 import org.cqfn.save.core.plugin.Plugin
 
 import okio.Path.Companion.toPath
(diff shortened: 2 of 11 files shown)


save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -21,6 +25,103 @@ expect class ProcessBuilder() {
...
+class ProcessBuilderInternal {
+    /**
+     * Typealias
+     */
+    val fs = FileSystem.SYSTEM
+
+    /**
+     * Temporary directory for stderr and stdout (popen can't separate streams, so we do it ourselves)
+     */
+    val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY / this::class.simpleName!!).also {
```
petertrr, line 40: We might run multiple `ProcessBuilder`s concurrently in the future

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -21,6 +25,103 @@ expect class ProcessBuilder() {
...
+     */
+    val stderrFile = tmpDir / "stderr.txt".also { logDebug("Created file for stderr of ProcessBuilder in: $tmpDir") }
+
+    /**
+     *  Read data from stdout file, we will use it in [ExecutionResult]
+     *
+     * @return string containing stdout
+     */
+    fun getStdout(): List<String> {
+        val stdout = fs.read(stdoutFile) {
```
petertrr, line 60: `readLines` from `FileUtils`?

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -21,6 +25,103 @@ expect class ProcessBuilder() {
...
+
+    /**
+     * Modify execution command for popen,
+     * stderr will be redirected to tmp file
+     *
+     * @param command raw command
+     * @return command with redirection of stderr to tmp file
+     */
+    fun prepareCmd(command: List<String>): String {
+        val cmd = command.joinToString(" ") + " 2>$stderrFile"
```
petertrr, line 86: What if command already contains stderr redirection? I think we should at least warn the user in this case.

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -21,6 +25,103 @@ expect class ProcessBuilder() {
...
+            return ExecutionResult(status, emptyList(), stderr)
+        }
+        redirectTo?.let {
+            fs.write(redirectTo) {
+                write(stdout.encodeToByteArray())
+            }
+        }
+            ?: run {
+                logDebug("Execution output:\n$stdout")
+            }
```
petertrr, line 120:
```suggestion
            ?: logDebug("Execution output:\n$stdout")
```
`run` is redundant here

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -21,6 +25,103 @@ expect class ProcessBuilder() {
...
+        return ExecutionResult(0, stdout.split("\n"), emptyList())
```
petertrr, line 121:
```suggestion
        return ExecutionResult(status, stdout.split("\n"), emptyList())
```
?

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -21,6 +25,103 @@ expect class ProcessBuilder() {
...
+        stdout: String,
+        status: Int,
+        redirectTo: Path?): ExecutionResult {
+        if (status == -1) {
+            fs.deleteRecursively(tmpDir)
+            error("Couldn't close the pipe, exit status: $status")
+        }
+        val stderr = getStderr()
+        fs.deleteRecursively(tmpDir)
+        if (stderr.isNotEmpty()) {
```
petertrr, line 109: I don't think we need a separate branch for this. For a process it is quite natural to write some data to stdout and some to stderr, let's give `ProcessBuilder`'s caller an opportunity to decide how to handle this.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
```diff
@@ -19,7 +19,7 @@ data class FixPluginConfig(
 ) : PluginConfig {
     init {
         require(inPlace || destinationFileSuffix != null) {
-            "Plugin ${FixPlugin::class.simpleName} should either be configured with inPlace=true or have destinationFileSuffix"
+            "Plugin ${FixPlugin::class.simpleName} should either be configured with inPlace = true or have destinationFileSuffix"
```
petertrr, line 22: I don't think spaces improve readability here

save-core/src/jvmMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -12,17 +16,33 @@ actual class ProcessBuilder {
     private val pb = ProcessBuilder()
 
     actual fun exec(command: List<String>, redirectTo: Path?): ExecutionResult {
+        val common = ProcessBuilderInternal()
```
petertrr, line 19: Can it be a property of expect class?

petertrr (changes requested):

petertrr:
I suggest also to add logging in the following places:
* `Save.performAnalysis`: which plugins are detected, which plugins start and complete execution

save-core/src/jvmMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: Expected property can't be initialized. What about creation `val common` without initialization - I don't know how reasonable is this. The code will be the same in `exec` function, instead of word `val`

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
kgevorkyan: Ok

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: I left separate branch for warning, but remove the extra `return`. Now function return stdout & stderr at the same time.

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: Yes, didn't notice. Thanks

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: Done

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: bash allows only one redirection per file descriptor.
If multiple redirections will be provided, they will processed from left to right, and the last one will take effect.
So we can warn the user, and don't really care about his redirection, since only our will be applied.

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: Applied

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: Fixed

save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
```diff
@@ -40,10 +46,13 @@ fun SaveProperties.validate(): SaveProperties {
...
     // getting configuration from command-line arguments
     val configFromCli = SaveProperties(args)
+    logDebug("Properties after parsed command line args:\n${configFromCli.getFields()}")
     // reading configuration from the properties file
     val configFromPropertiesFile = readPropertiesFile(configFromCli.propertiesFile)
-    // merging two configurations into signle [SaveProperties] class with a priority to command line arguments
-    return configFromCli.mergeConfigWithPriorityToThis(configFromPropertiesFile).validate()
+    // merging two configurations into single [SaveProperties] class with a priority to command line arguments
+    val mergedProperties = configFromCli.mergeConfigWithPriorityToThis(configFromPropertiesFile)
+    logInfo("Summary properties:\n${mergedProperties.getFields()}")
```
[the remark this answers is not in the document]
kgevorkyan: Done

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -37,7 +41,8 @@ class ProcessBuilderInternal {
     /**
      * Temporary directory for stderr and stdout (popen can't separate streams, so we do it ourselves)
      */
-    val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY / this::class.simpleName!!).also {
+    val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY /
+            (this::class.simpleName!! + "_" + Clock.System.now().toString().substringAfterLast(".")).toPath()).also {
```
petertrr, line 45:
```suggestion
            (this::class.simpleName!! + "_" + Clock.System.now().epochSeconds).toPath()).also {
```
Maybe, just with timestamp?

save-core/src/jvmMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
petertrr: I think we should actually do vice-versa: have ordinary `class ProcessBuilder` in common sources along with `expect class ProcessBuilderInternal`; `ProcessBuilder` can have a property of type `ProcessBuilderInternal` and call it's methods inside common logic of `exec`

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
kgevorkyan: Done

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -22,8 +23,11 @@ class Save(
         requireNotNull(testConfig) { "Provided path ${saveProperties.testConfig} doesn't correspond to a valid save.toml file" }
 
         val plugins: List<Plugin> = emptyList()  // todo: discover plugins (from configuration blocks in TestSuiteConfig?)
+        logInfo("Discovered plugins: $plugins")
         plugins.forEach {
+            logInfo("Execute plugin: ${it::class::simpleName}")
             it.execute(saveProperties, testConfig)
+            logInfo("${it::class::simpleName} successfully executed!")
```
petertrr, line 30:
```suggestion
            logInfo("${it::class.simpleName} successfully executed!")
```
I believe class name is a property, not a reference

petertrr (changes requested):

PR #41, opened 2021-04-21
Refactor implementation of ProcessBuilder

kgevorkyan:

PR #49, opened 2021-04-28
Support windows server and ubuntu in CI

kgevorkyan:


.github/workflows/build_and_test.yml
```diff
@@ -51,3 +51,48 @@ jobs:
         with:
           flags: unittests
           fail_ci_if_error: true # optional (default = false)
+
+  build_and_test:
```
petertrr, line 55: I think we can unify all jobs into one and just add a condition to `upload_code_coverage` step thatit should be executed only on Linux (but for now it should be still not active).

save-core/src/jvmMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -18,6 +18,7 @@ actual class ProcessBuilder {
     actual fun exec(command: List<String>, redirectTo: Path?): ExecutionResult {
         val common = ProcessBuilderInternal()
 
+        // FixMe: move it to the common isWindows method
```
petertrr, line 21: You already have this method, you should use it here

save-plugins/fix-plugin/src/commonTest/kotlin/org/cqfn/save/plugins/fix/FixPluginTest.kt
```diff
@@ -116,19 +128,29 @@ class FixPluginTest {
         }
         val expectedFile = fs.createFile(tmpDir / "Test3Expected.java")
         fs.write(expectedFile) {
-            write("Expected file".encodeToByteArray())
+            if (isCurrentOsWindows()) {
+                write("Expected file \n".encodeToByteArray())
```
petertrr, line 132: So we still have a problem with this trailing space? Even if we are using cmd again, and not powershell?

petertrr (commented):
Generall LGTM, please take a look at a couple of questions

save-plugins/fix-plugin/src/commonTest/kotlin/org/cqfn/save/plugins/fix/FixPluginTest.kt
kgevorkyan: Yes, I found, that this problem initially appeared in CMD, and repeated in PowerShell for Windows.

In https://github.com/cqfn/save/pull/41, in tests of ProcessBuilder, I've encountered with same problems.
For now, I don't now some other workarounds

petertrr: OK, let's keep it like this. But please don't repeat line with `write` twice; introduce a viraibale `textPostfix` and call `write("Exceptec file$textPostfix")` to reduce code duplication

petertrr (approved):

.github/workflows/build_and_test.yml
```diff
@@ -9,7 +9,10 @@ on:
 jobs:
   build_and_test_with_code_coverage:
     name: Build, test and upload code coverage
```
petertrr, line 11:
```suggestion
    name: Build and test
```
Let's call it simply `Build and test`, also I'll update required checks (they are set by name) so that this PR can be merged

PR #56, opened 2021-04-30
Introduce API for reporters

### LINKED ISSUE #40: Design and implement API for reporters
With outputs of different format. Should be done after #38 

Reporter is like a test events listener. In #38 I have proposed lazy sequence of test results being returned from plugins. With this we can call reporter on every new value occurring in the sequence.
Reporter should also have special methods `beforeAll`, `afterAll`; optionally - somthing like `onPluginInitialization`, `onPluginExecutionStart`, `onPluginExecutionEnd`, `onInternalError`.

Things for inspiration: [junit](https://junit.org/junit5/docs/5.0.3/api/org/junit/platform/launcher/TestExecutionListener.html), `Reporter` from ktlint and detekt

petertrr:
### What's done:
* Added interface Reporter
* Supported it in Save.kt
* Added StdoutSink 
* Added save.properties in examples

Closes #40 

### TODO:
* [x] ~Support output to stdout and/or file~ Output stream will be provided based on property `resultOutputType`
* [x] Add implementation of simple plain text reporter


kgevorkyan (approved):
lgtm

save-common/src/commonMain/kotlin/org/cqfn/save/core/reporter/Reporter.kt
```diff
@@ -0,0 +1,115 @@
...
+import okio.BufferedSink
+
+/**
+ * A base interface for all reporters, that are capable of formatting results of SAVE execution
+ */
+interface Reporter {
+    /**
+     * A type of this reporter
+     */
+    val type: ReportType
```
akuleshov7, line 17: for config plugins you didn't want to have type

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -72,11 +91,26 @@ class Save(
...
+            reporter.onPluginInitialization(plugin)
+        }
+
         plugins.forEach { plugin ->
             logInfo("Execute plugin: ${plugin::class.simpleName}")
-            plugin.execute().forEach(this::handleResult)
+            reporter.onPluginExecutionStart(plugin)
+            try {
+                plugin.execute()
+                    .onEach { event -> reporter.onEvent(event) }
```
akuleshov7, line 103: it?

save-common/src/commonMain/kotlin/org/cqfn/save/core/reporter/Reporter.kt
petertrr: Well, for reporters it's natural to have an explicit configuration, and not in toml, but rather in properties or even CLI. So I think explicitly listing types of reports is more appropriate, than hard-coding types of plugins

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
petertrr: I wanted to be more explicit, it could be shortened even to `.onEach(reporter::onEvent)`

akuleshov7 (approved):

PR #62, opened 2021-05-12
Fix cli release upload to github

### LINKED ISSUE #58: fix cli release upload to github
In #51 we had an issue related to Save upload to github. Need to fix it, so everything will be automated

Cheshiriks:
What's done:
* add linkReleaseExecutableMultiplatform task

close #58

### DIFF
--- a/save-cli/build.gradle.kts
+++ b/save-cli/build.gradle.kts
@@ -15,10 +15,12 @@ kotlin {
     val os = getCurrentOperatingSystem()
     val saveTarget = listOf(macosX64(), linuxX64(), mingwX64())
 
-    configure(saveTarget) {
-        binaries {
-            executable {
-                entryPoint = "org.cqfn.save.cli.main"
+    saveTarget.forEach {
+        configure(listOf(it)) {
+            binaries {
+                executable("save-${it.name}") {
+                    entryPoint = "org.cqfn.save.cli.main"
+                }
             }
         }
     }
--- a/save-cli/build.gradle.kts
+++ b/save-cli/build.gradle.kts
@@ -54,6 +54,14 @@ kotlin {
             }
         }
     }
+
+    project.tasks.register("linkReleaseExecutableMultiplatform") {
+        when (saveTarget.name) {
+            "linuxX64" -> dependsOn(tasks.getByName("linkReleaseExecutableLinuxX64"))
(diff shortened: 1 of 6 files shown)


save-cli/build.gradle.kts
```diff
@@ -54,6 +54,14 @@ kotlin {
             }
         }
     }
+
+    project.tasks.register("linkReleaseExecutableMultiplatform") {
+        when (saveTarget.name) {
+            "linuxX64" -> dependsOn(tasks.getByName("linkReleaseExecutableLinuxX64"))
+            "mingwX64" -> dependsOn(tasks.getByName("linkReleaseExecutableMingwX64"))
+            "macosX64" -> dependsOn(tasks.getByName("linkReleaseExecutableMacosX64"))
+        }
```
petertrr, line 63:
```suggestion
        dependsOn(tasks.getByName("linkReleaseExecutable${saveTarget.name.capitalize()}"))
```
Maybe, like this?

save-core/build.gradle.kts
```diff
@@ -16,15 +15,11 @@ kotlin {
             }
         }
     }
-    val os = getCurrentOperatingSystem()
-    // Create a target for the host platform.
-    val hostTarget = when {
```
petertrr, line 21: This should be changed for plugins as well, it affects all artifacts that we are deploying

petertrr (changes requested):

save-core/build.gradle.kts
Cheshiriks: Fixed

petertrr: And `warn-plugin`?

Cheshiriks: Done

petertrr (commented):
Maybe, action `download-artifact` creates a directory for artifacts, and because of that `upload-asset` skips them?

petertrr (commented):
We should also add a constraint to release workflow: start build only if git working tree is clean. It can be done in a separate issue, though

Cheshiriks:
> Maybe, action `download-artifact` creates a directory for artifacts, and because of that `upload-asset` skips them?

But we ourselves create a directory for loading artifacts in _download-artifact_ - tmpFolder. And then we try to upload them from it in _upload-release-asset_.

petertrr (approved):
lgtm

PR #63, opened 2021-05-12
Common configuration for test resources

### LINKED ISSUE #43: Add common configuration for test resources
Either pattern in `[general]` config or common setting for all plugins. With latter approach we will still be able to treat, e.g., pair of `*Test` and `*Expected` files as a single test.

And this API can be used in preprocessor in save-cloud to dsicover all tests for any generic cofnig

petertrr:
### What's done:
* Changed PluginConfig, added `discoverTestFiles` method to Plugin
* Updated fix-plugin and warn-plugin
* Fixed tests

Closes #43


save-core/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -16,4 +18,12 @@ interface Plugin {
...
      */
     fun execute(saveProperties: SaveProperties, testConfig: TestConfig): Sequence<TestResult>
+
+    /**
+     * Discover groups of resource files which will be used to run tests.
+     *
+     * @param root root [Path], from where discovering should be started
+     * @return a sequence of files, grouped by test
+     */
+    fun discoverTestFiles(root: Path): Sequence<List<Path>>
```
kgevorkyan, line 28: It seems, that return value actually is `Sequence<Pair<Path>>`, right?

May be we can explicitly specify this?

petertrr: No, `Pair` and `List` are not related, so they couldn't be cast to each other. We have a list with single element for `warn-plugin` and a list of two elements for `fix-plugin`, so to keep the API more general I've put a list here

kgevorkyan: Ok, just saw something like `.map { it.first() to it.last() }`, so I thought, that we have the `Pair`. Also previously we had `discoverFilePairs`

save-core/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -14,13 +15,25 @@ data class TestConfig(
...
      */
     val childConfigs: MutableList<TestConfig> = mutableListOf()
 
+    /**
+     * Directory containing [location] of this config
+     */
+    val directory: Path = location.parent!!
+
+    init {
+        require(!fs.metadata(location).isDirectory) {
```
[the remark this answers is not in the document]
petertrr: I think we can, I will change

kgevorkyan (approved):

akuleshov7 (approved):
lgtm

PR #66, opened 2021-05-14
Handle extra whitespaces and newlines in Windows `echo`

kgevorkyan:
### What's done:
* Provide additional function which will wrap `echo` command on Windows in the following manner `echo | set /p="command"`
* Add tests for this wrap algorithm
* Fix old tests, which uses `echo` commands

### DIFF
--- a/save-core/src/nativeTest/kotlin/org/cqfn/save/core/utils/ProcessBuilderInternalTest.kt
+++ b/save-core/src/nativeTest/kotlin/org/cqfn/save/core/utils/ProcessBuilderInternalTest.kt
@@ -10,12 +10,12 @@ class ProcessBuilderInternalTest {
     @Test
     fun `check stdout with redirection`() {
         val actualResult = processBuilder.exec("echo something >/dev/null", null)
-        val expectedCode = 0
-        val expectedStdout = "something"
+        val expectedCode = if (isCurrentOsWindows()) 1 else 0
+        val expectedStdout = listOf("something")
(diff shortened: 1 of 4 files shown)


save-core/src/nativeTest/kotlin/org/cqfn/save/core/utils/ProcessBuilderInternalTest.kt
```diff
@@ -10,12 +10,12 @@ class ProcessBuilderInternalTest {
     @Test
     fun `check stdout with redirection`() {
         val actualResult = processBuilder.exec("echo something >/dev/null", null)
-        val expectedCode = 0
-        val expectedStdout = "something"
+        val expectedCode = if (isCurrentOsWindows()) 1 else 0
```
petertrr, line 13: Why does it return 1 now? Wouldn't it be confusing?

kgevorkyan: You can look at [this](https://stackoverflow.com/questions/7105433/windows-batch-echo-without-new-line/11336754#11336754)

For now, exit codes are checked only by our tests, also I I don't think, that users will use `echo` commands regularly in SAVE. It's a good question, but I don't think, that to use some dummName (advise  in stackoverflow) is good approach too, so I left it as-is

petertrr (approved):
lgtm

PR #69, opened 2021-05-14
Introduce save-common subproject 

petertrr:
### What's done:
* Moved files from save-core


save-common/build.gradle.kts
```diff
@@ -0,0 +1,64 @@
...
+
+kotlin {
+    jvm {
+        compilations.all {
+            kotlinOptions {
+                jvmTarget = "11"
+            }
+        }
+    }
+    val hostTarget = listOf(linuxX64(), mingwX64(), macosX64())
```
akuleshov7, line 17: macos?

akuleshov7 (approved):
lgtm

save-common/build.gradle.kts
petertrr: Yeah, commented it out for now

PR #72, opened 2021-05-19
Distinguish resources among different SAVE configs

petertrr:
### What's done:
* Introduced default method `resourceDirectories` in `Plugin` to get resource directories managed by this config
* Added `suiteName` property to `GeneralConfig`

### TODO:
* Call this method in plugin interface

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
@@ -87,7 +87,7 @@ fun FileSystem.readLines(path: Path): List<String> = this.read(path) {
  * |   |-- directory23
  * |   |   `-- file33
  * ```
- * If predicate returns `true` when `Path` is a directory which contains only other directories, then `directory11` will be filtered, as well as
+ * If predicate returns `true` when `Path` is a directory which contains only other directories or is empty, then `directory11` will be filtered, as well as
  * `directory21`, which is empty, but is a descendant of already filtered out one, and `directory 23`, which is not empty.
  * So, the result would be `sequence(directory12, directory13)`
  *
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
@@ -77,30 +77,33 @@ fun FileSystem.readLines(path: Path): List<String> = this.read(path) {
  * Returns a sequence of underlying directories, filtering on every level by [directoryPredicate].
  * Example:
  * ```
- * | directory1
- * |_ file1
- * | | directory11
- * | |_ file2
- * | |_ directory21
- * | | directory12
- * | | directory13
- * | |_ directory23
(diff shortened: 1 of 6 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
```diff
@@ -72,3 +72,35 @@ fun FileSystem.createFile(path: Path): Path {
...
+ * ```
+ * | directory1
+ * |_ file1
+ * | | directory11
+ * | |_ file2
+ * | |_ directory21
+ * | | directory12
+ * | | directory13
+ * | |_ directory23
+ * | | |_ file33
```
kgevorkyan, line 88:
```suggestion
 * |____directory1
 * |       |____file1
 * |       |____directory11
 * |               |____file2
 * |               |____directory21
 * |       |____directory12
 * |       |____directory13
 * |       |____directory23
 * |               |____file33
```

petertrr: I've changed it to more `tree`-like output, also added a test with the same example.

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -24,4 +27,17 @@ interface Plugin {
...
+
+    /**
+     * Returns a sequence of directories, where resources for this plugin may be located.
+     * This takes into account, that if underlying directory contains it's own SAVE config,
+     * then this plugin shouldn't touch these resources; it should be done by plugins from that config.
+     *
+     * @param fs a [FileSystem] which is used to traverse the directory hierarchy
+     * @return a sequence of directories possibly containing this plugin's test resources
+     */
+    fun Path.resourceDirectories(fs: FileSystem = FileSystem.SYSTEM): Sequence<Path> = findDescendantDirectoriesBy(true) { file ->
```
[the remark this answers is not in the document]
petertrr: I don't really like the idea of a single global object, that we are using to get access to the filesystem (`FileSystem.SYSTEM`). For example, there is `okio-fakefilesystem`, which could be used to run tests with mocked in-memory filesystem, but there is no convenient way to inject it everywhere. I still haven't quite figured out, how to do it better, so I try to at least assign this object to some variables

kgevorkyan (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -11,5 +11,9 @@ interface PluginConfig
 
 /**
  * General configuration for test suite.
+ * @property suiteName name of the test suite
  */
-class GeneralConfig : PluginConfig
+@Suppress("INLINE_CLASS_CAN_BE_USED")
+class GeneralConfig(
+    val suiteName: String,
```
akuleshov7, line 18: will have conflicts here

akuleshov7 (approved):
discussed offline, looks good

PR #77, opened 2021-05-20
Check if working tree is clean in release build

petertrr:
### What's done:
* Get jgit from grgit to get git status
* Check if working tree is clean if proeprty `release` is present

### DIFF
--- a/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/VersioningConfiguration.kt
+++ b/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/VersioningConfiguration.kt
@@ -6,19 +6,16 @@
 
 package org.cqfn.save.buildutils
 
+import org.ajoberstar.grgit.Grgit
 import org.ajoberstar.reckon.gradle.ReckonExtension
 import org.ajoberstar.reckon.gradle.ReckonPlugin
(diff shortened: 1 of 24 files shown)


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/VersioningConfiguration.kt
```diff
@@ -2,20 +2,44 @@
...
+    require(this == rootProject) { "Versioning should be configured for the root project" }
+
+    apply<GitVersionPlugin>()
+
+    val versionDetails: Closure<VersionDetails> by extra
+    val details = versionDetails.invoke()
+
+    allprojects {
+        version = details.version.let {
+            if (it.startsWith("v")) it.trim('v') else it
```
akuleshov7, line 35: why not to check that versioning is correct?
check for the semver format

akuleshov7 (approved):
lgtm

buildSrc/src/main/kotlin/org/cqfn/save/buildutils/VersioningConfiguration.kt
petertrr: Added a pattern, requiring that tags start with `'v'`

PR #79, opened 2021-05-21
Created the logic for merging of save.toml configs

kgevorkyan:
Created the logic for merging of save.toml configs
### What's done:
* Provide logic for merging of nested save.toml configs
* Add tests
* Refactor ConfigDetector


save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
```diff
@@ -15,10 +15,6 @@ data class FixPluginConfig(
     val execCmd: String,
     val destinationFileSuffix: String? = null,
 ) : PluginConfig {
-    init {
```
kgevorkyan, line 18: This will be reworked in https://github.com/cqfn/save/issues/86

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/MergeConfigs.kt
```diff
@@ -0,0 +1,112 @@
...
+        logDebug("Merging ${parent.location} with ${child.location}")
+        val parentConfig = parent.pluginConfigs
+        val childConfig = child.pluginConfigs
+        // Create the list of corresponding configs, if one of them will be null -> list will contain only one element,
+        // which we apply as final config, otherwise we will merge configs
+        val generalConfigs = listOfNotNull(
+            parentConfig.filterIsInstance<GeneralConfig>().firstOrNull(),
+            childConfig.filterIsInstance<GeneralConfig>().firstOrNull()
+        )
+        val newGeneralConfig: GeneralConfig? = if (generalConfigs.size != 2) generalConfigs.firstOrNull() else mergeGeneralConfigs(generalConfigs.first(), generalConfigs.last())
```
petertrr, line 67:
```suggestion
        val newGeneralConfig: GeneralConfig? = generalConfigs.singleOrNull() ?: mergeGeneralConfigs(generalConfigs.first(), generalConfigs.last())
```

petertrr (changes requested):

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -60,7 +61,7 @@ class Save(
         val testConfig = ConfigDetector().configFromFile(fullPathToConfig)
 
         val plugins: List<Plugin> = discoverPluginsAndUpdateTestConfig(testConfig)
-            // FixMe: Kirill should add merging of configs here
+            .also { MergeConfigs().merge(testConfig) }
```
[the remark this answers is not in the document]
petertrr: Oh, yes, you are merging them in place by mutating list of plugins in every config, I see

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/MergeConfigs.kt
```diff
@@ -0,0 +1,112 @@
...
+     */
+    fun merge(testConfig: TestConfig) {
+        logDebug("Start merging configs for ${testConfig.location}")
+        val parentConfigs = collectParentConfigs(testConfig)
+        mergeConfigList(parentConfigs)
+        mergeChildConfigs(testConfig)
+    }
+
+    // Create the list of parent configs
+    private fun collectParentConfigs(testConfig: TestConfig): MutableList<TestConfig> {
```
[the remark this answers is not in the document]
kgevorkyan: Yes, thank you

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/MergeConfigs.kt
kgevorkyan: Basically, I need to consider three cases:
1) If list is empty -> return null
2) If `list.size == 1` -> return single element
3) If `list.size == 2` -> return merge elements
-----------------------
But your suggestion will give the following:
1) If list is empty -> singleOrNull returns null -> mergeConfgis -- ERROR
2) If `list.size == 1` -> return single element -- OK
3) If `list.size == 2` -> return merge elements -> singleOrNull returns null -> mergeConfgis -- OK

kgevorkyan: I think, the reason was in unclear comment before this code, I provided more detailed explanation

petertrr (approved):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
```diff
@@ -35,24 +35,13 @@ data class WarnPluginConfig(
     val execCmd: String,
     val warningsInputPattern: Regex,
     val warningsOutputPattern: Regex,
-    val warningTextHasLine: Boolean = true,
-    val warningTextHasColumn: Boolean = true,
+    val warningTextHasLine: Boolean? = null,
```
akuleshov7, line 38: Kirill, as you remember, I have a short memory about these solutions, that's why:
1) please write some comment about it - why we have instantiated it with null
2) looks like you have broken the logic for the default value with these arguments, please check

kgevorkyan: 1) In process of merging we can met following situation:
Parent config have field `warningTextHasLine = false`
Child config have no value for this field in `toml` config, but since it's have default value `warningTextHasLine = true`,
as the result we will apply this `true`, of which there was no information in config at all.
We can't detect, whether we deal with default value in child config, or it  was provided by user.

2) I think we don't support default values at all, according 1).

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
```diff
@@ -35,24 +35,13 @@ data class WarnPluginConfig(
...
+    val warningTextHasColumn: Boolean? = null,
     val lineCaptureGroup: Int?,
     val columnCaptureGroup: Int?,
     val messageCaptureGroup: Int,
     val exactWarningsMatch: Boolean = true,
```
akuleshov7, line 43: fix this also!

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/MergeConfigs.kt
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * A class that is capable for merging inherited configurations
+ */
+class MergeConfigs {
+    /**
+     * Merge parent configurations with current and prolong it for all child configs
+     *
+     * @param testConfig - testing SAVE config (save.toml) which should be merged
+     */
+    fun merge(testConfig: TestConfig) {
```
akuleshov7, line 18: move this logic to TestConfig

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/MergeConfigs.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    private fun mergeGeneralConfigs(parentConfig: GeneralConfig, childConfig: GeneralConfig) = GeneralConfig(
+        childConfig.tags ?: parentConfig.tags,
+        childConfig.description ?: parentConfig.description,
+        childConfig.suiteName ?: parentConfig.suiteName,
+        childConfig.excludedTests ?: parentConfig.excludedTests,
+        childConfig.includedTests ?: parentConfig.includedTests,
+    )
+
+    private fun mergeWarnConfigs(parentConfig: WarnPluginConfig, childConfig: WarnPluginConfig) = WarnPluginConfig(
```
akuleshov7, line 85: this should go to plugin

kgevorkyan: Done

save-core/src/commonMain/kotlin/org/cqfn/save/core/files/MergeConfigs.kt
kgevorkyan: Moved

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -65,6 +71,56 @@ data class TestConfig(
...
+        }
+    }
+
+    private fun mergeChildConfigWithParent(parent: TestConfig) {
+        logDebug("Merging ${parent.location} with ${this.location}")
+        val parentPluginConfigs = parent.pluginConfigs
+        val childPluginConfigs = this.pluginConfigs
+
+        // Going through parent configs and:
+        // If some config is absent in parent, but exists is child, leave it as it is
```
petertrr, line 111: How do we keep these configs in `result`? It looks like they are lost

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -9,10 +9,31 @@ import kotlinx.serialization.Serializable
...
-interface PluginConfig
+interface PluginConfig<T : PluginConfig<T>> {
+    /**
+     * Create new config by merging of [this] and corresponding config from [childConfig]
+     *
+     * @param childConfig list of child configs, which will be filtered for merging
+     * @return new config
+     */
+    @Suppress("TYPE_ALIAS")
+    fun createNewPluginConfig(childConfig: MutableList<PluginConfig<*>>): T
```
petertrr, line 20:
```suggestion
    fun createNewPluginConfig(childConfig: List<PluginConfig<*>>): T
```

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -65,6 +71,56 @@ data class TestConfig(
...
+        // If some config is absent in child, but exists in parent, just take it from parent
+        // Otherwise we will merge configs
+        val result: MutableList<PluginConfig<*>> = mutableListOf()
+        for (config in parentPluginConfigs) {
+            val newPluginConfig = config.createNewPluginConfig(childPluginConfigs)
```
petertrr, line 116: Shouldn't it be vice versa? Iterating over `childPluginConfigs` and merging them with `parentPluginConfigs`?

petertrr (changes requested):

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
kgevorkyan: Yes, thank you! Improved the logic

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
kgevorkyan: For now, we modify childConfig in this method, so I left mutable list

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
kgevorkyan: No, we need to insert/merged only those configs, which are exists in parent, so we iterate over them
Renamed the method to more suitable

petertrr (approved):

akuleshov7 (approved):
I tried to do something with this ugly generic, but it looks like it is really the only variant for compile-time, lgtm

PR #82, opened 2021-05-24
Implement logic for ignore-save-comments parameter in warn-plugin

### LINKED ISSUE #45: Implement logic for ignore-save-comments parameter in warn-plugin
Continuation of #17. There are several ways to do this, which need careful consideration:
* Adjust line numbers from expected warnings based on lines with technical comments. Drawback: there are some warnings that are triggered on the whole file and they usually have line:col==1:1, which will be hard to adjust. We could suggest setting 0:1 for such warnings in comments.
* Copy all test files with technical comment filtered out and run tool on them. Drawback: we can't adjust the tool, because we operate with `execCmd` as provided. We could write filtered content into `execCmd`s stdin, if it expects `stdin`. We could also ask users to configure `execCmd` appropriately in this mode.

Cheshiriks:
### What's done:
* Added General config - ignoreSaveComments
* Added logic for ignore warning comments
* Added test

Closes #45

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
@@ -6,8 +6,6 @@ package org.cqfn.save.core.plugin
 
 import kotlinx.serialization.Serializable
 
-import kotlin.jvm.JvmInline
-
 /**
  * Core interface for plugin configuration
  */
@@ -23,7 +21,6 @@ interface PluginConfig
  * @property ignoreSaveComments if true then ignore warning comments
  */
 @Serializable
-@JvmInline
 data class GeneralConfig(
     val tags: String,
     val description: String,
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
@@ -6,6 +6,8 @@ package org.cqfn.save.core.plugin
 
 import kotlinx.serialization.Serializable
 
+import kotlin.jvm.JvmInline
+
 /**
  * Core interface for plugin configuration
  */
@@ -18,12 +20,15 @@ interface PluginConfig
  * @property suiteName
  * @property excludedTests FixMe: after ktoml will support lists we should change it
  * @property includedTests FixMe: after ktoml will support lists we should change it
+ * @property ignoreSaveComments if true then ignore warning comments
  */
 @Serializable
+@JvmInline
 data class GeneralConfig(
     val tags: String,
     val description: String,
     val suiteName: String,
     val excludedTests: String = "",
     val includedTests: String = "",
+    val ignoreSaveComments: Boolean = false,
 ) : PluginConfig
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
@@ -19,6 +19,11 @@ enum class CurrentOs {
 @Suppress("VARIABLE_NAME_INCORRECT_FORMAT", "ConstructorParameterNaming")
 expect class AtomicInt(value_: Int) {
     /**
+     * @param value the value being held by this class.
+     */
+    var value: Int
+
+    /**
      * @param delta increments the value_ by delta
      * @return the new value
      */
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
@@ -17,11 +17,11 @@ enum class CurrentOs {
  * Atomic values
  */
 @Suppress("VARIABLE_NAME_INCORRECT_FORMAT", "ConstructorParameterNaming")
-expect class AtomicInt(value_: Int) {
+expect class AtomicInt(value: Int) {
     /**
-     * @param value the value being held by this class.
+     * @return value
      */
-    var value: Int
+    fun get(): Int
 
     /**
      * @param delta increments the value_ by delta
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
@@ -16,7 +16,6 @@ enum class CurrentOs {
 /**
  * Atomic values
  */
-@Suppress("VARIABLE_NAME_INCORRECT_FORMAT", "ConstructorParameterNaming")
 expect class AtomicInt(value: Int) {
     /**
      * @return value
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
@@ -14,6 +14,18 @@ enum class CurrentOs {
 }
 
 /**
+ * Atomic values
+ */
+@Suppress("VARIABLE_NAME_INCORRECT_FORMAT", "ConstructorParameterNaming")
+expect class AtomicInt(value_: Int) {
+    /**
+     * @param delta increments the value_ by delta
+     * @return the new value
+     */
+    fun addAndGet(delta: Int): Int
+}
+
+/**
  * Get type of current OS
  *
  * @return type of current OS
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
@@ -17,6 +17,7 @@ import org.cqfn.save.plugin.warn.utils.extractWarning
 
 import okio.FileSystem
 import okio.Path
+import kotlin.jvm.JvmStatic
 
 private typealias LineColumn = Pair<Int, Int>
 
@@ -45,7 +46,6 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
         .filter { it.isNotEmpty() }
 
     @Suppress(
-        "UnusedPrivateMember",
         "TOO_LONG_FUNCTION",
         "SAY_NO_TO_VAR")
     private fun handleTestFile(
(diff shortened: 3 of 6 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -26,4 +27,5 @@ data class GeneralConfig(
     val suiteName: String,
     val excludedTests: String = "",
     val includedTests: String = "",
+    val ignoreSaveComments: Boolean = false,
```
petertrr, line 30: @kgevorkyan after all, are we making these fields nullable? Or you will handle it in #79 ?

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -42,10 +45,14 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
         .asSequence()
         .filterNot { it.isEmpty() }
 
-    @Suppress("UnusedPrivateMember")
+    @Suppress(
+        "UnusedPrivateMember",
```
petertrr, line 49: Why did we need this suppress?)

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -65,8 +72,15 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
...
             .mapValues { it.value.sortedBy { it.message } }
-        // todo: create a temp file with technical comments removed and feed it to the tool
-        val executionResult = pb.exec(warnPluginConfig.execCmd, null)
+
+        var execCmd: String = warnPluginConfig.execCmd
+
+        if (generalConfig.ignoreSaveComments) {
+            createTestFile(path)
+            execCmd = warnPluginConfig.execCmd + " test_file"
+        }
```
petertrr, line 81:
```suggestion

        val execCmd = if (generalConfig.ignoreSaveComments) {
            createTestFile(path)
            warnPluginConfig.execCmd + " test_file"
        } else {
            warnPluginConfig.execCmd
        }
```
I don't like mutable properties :)

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -65,8 +72,15 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
...
             }
             .mapValues { it.value.sortedBy { it.message } }
-        // todo: create a temp file with technical comments removed and feed it to the tool
-        val executionResult = pb.exec(warnPluginConfig.execCmd, null)
+
+        var execCmd: String = warnPluginConfig.execCmd
+
+        if (generalConfig.ignoreSaveComments) {
+            createTestFile(path)
+            execCmd = warnPluginConfig.execCmd + " test_file"
```
petertrr, line 80: `test_file` should be extracted to a variable

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -81,16 +95,32 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
...
+        val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY / WarnPlugin::class.simpleName!!)
+
+        if (fs.exists(tmpDir)) {
+            fs.deleteRecursively(tmpDir)
+        }
+        fs.createDirectory(tmpDir)
+
+        fs.write(fs.createFile(tmpDir / "test_file")) {
+            fs.readLines(path).forEach {
+                if (!it.contains("// ;warn:")) {
```
petertrr, line 111: Do we have a pattern for this somewhere? There could be languages (like python) where `//` isn't a comment sign

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -81,16 +95,32 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
...
+     */
+    internal fun createTestFile(path: Path) {
+        val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY / WarnPlugin::class.simpleName!!)
+
+        if (fs.exists(tmpDir)) {
+            fs.deleteRecursively(tmpDir)
+        }
+        fs.createDirectory(tmpDir)
+
+        fs.write(fs.createFile(tmpDir / "test_file")) {
```
petertrr, line 109: In multithreaded mode plugin will break, because files have same name

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -65,8 +72,15 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
                 }
             }
             .mapValues { it.value.sortedBy { it.message } }
-        // todo: create a temp file with technical comments removed and feed it to the tool
-        val executionResult = pb.exec(warnPluginConfig.execCmd, null)
+
+        var execCmd: String = warnPluginConfig.execCmd
```
petertrr, line 76: We have discussed, that we should append file name to the cmd anyway (like we discussed for #88). I think we should do it here too

petertrr (changes requested):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
kgevorkyan: Yes, we decide to make such options nullable for now. I can handle it, if this PR will be merged first. Or @Cheshiriks can change it by his own

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
Cheshiriks: fixed

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -28,8 +30,9 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
 
     override fun execute(): Sequence<TestResult> {
         val warnPluginConfig = testConfig.pluginConfigs.filterIsInstance<WarnPluginConfig>().single()
+        val generalConfig = testConfig.pluginConfigs.filterIsInstance<GeneralConfig>().single()
```
petertrr, line 34: Let's make general config optional here, and if it is null, then the default value for `ignore...` parameter is used

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -81,16 +95,35 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
...
 
+    /**
+     * @param path
+     * @return name of the temporary file
+     */
+    internal fun createTestFile(path: Path): String {
+        val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY / WarnPlugin::class.simpleName!!)
+
+        if (fs.exists(tmpDir)) {
+            fs.deleteRecursively(tmpDir)
```
petertrr, line 106: This directory should be created per SAVE execution, because different threads running the same plugin will create files there. You could check if your counter is 0 and then delete, otherwise do nothing

petertrr (changes requested):
Also, please don't force-push after review, it's really hard to track what has been changed

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/PlatformUtils.kt
```diff
@@ -13,6 +13,23 @@ enum class CurrentOs {
     LINUX, MACOS, UNDEFINED, WINDOWS
 }
 
+/**
+ * Atomic values
+ */
+@Suppress("VARIABLE_NAME_INCORRECT_FORMAT", "ConstructorParameterNaming")
```
petertrr, line 19: Since we are not using typealias for native anymore, we can use proper variable naming. Also, what does diktat complain about here?

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -80,16 +94,37 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
...
+            fs.createDirectory(tmpDir)
+        } else if (!fs.exists(tmpDir)) {
+            fs.createDirectory(tmpDir)
+        }
+
+        val fileName = testFileName()
+        fs.write(fs.createFile(tmpDir / fileName)) {
+            fs.readLines(path).forEach {
+                if (!it.contains(";warn:")) {
```
petertrr, line 114:
```suggestion
                if (!warningsInputPattern.matches(it)) {
```
We have a configuration option for this. Also, please don't force-push after reviews.

petertrr (changes requested):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
Cheshiriks: Force-push must be done after the merge the latest changes from _main_

petertrr (approved):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
petertrr: Use `merge from main` instead of `rebase`

PR #85, opened 2021-05-25
Bugfix: fixed logic of Path.findAllFilesMatching

### LINKED ISSUE #83: Bug in findAllFilesMatching returning empty lists
Need to filter non-empty lists before returning

petertrr:
### What's done:
* Fixed logic
* Added test

Closes #83


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -40,7 +40,6 @@ class WarnPlugin(testConfig: TestConfig) : Plugin(testConfig) {
                 .filter { defaultResourceNamePattern.matches(it.name) }
         }
         .asSequence()
-        .filterNot { it.isEmpty() }
```
petertrr, line 43: It seems that it actually is not just a workaround (added by @akuleshov7) for a bug, but rather a feature: after we discover all directories for this config, some of them may be empty (i.e. without any test files), and we should filter them out here

akuleshov7: Haha true

akuleshov7 (approved):

PR #90, opened 2021-05-26
Configure snapshots publication

petertrr:
### What's done:
* Changed reckon plugin configuration
* Update release.yml
* Use latest checkout v2


akuleshov7 (approved):
lgtm

PR #97, opened 2021-05-27
Run individual tests

### LINKED ISSUE #93: Add logic for an execution of one single test
Now occasionally we can run tests only by config, need to think about the logic for running of one single test

petertrr:
### What's done:
* Changed Plugin API
* Added logic
* Updated readme

Closes #93


buildSrc/src/main/kotlin/org/cqfn/save/generation/Generation.kt
```diff
@@ -82,6 +82,22 @@ fun FunSpec.Builder.generateOptions(jsonObject: Map<String, Option>): FunSpec.Bu
     return this
 }
 
+/**
+ * Adds statement with vararg CLI argument for testFiles
+ *
+ * @return builder
+ */
+fun FunSpec.Builder.addTestsVararg(): FunSpec.Builder = apply {
```
kgevorkyan, line 90: May be we can create additional `if` in `FunSpec.Builder.generateOptions`, and if `it.key == testFiles`, than manually add 
```
.optional()
.vararg()
```
statements instead of this function? Also, seems you're forget about `fullName` for this option

petertrr: > Also, seems you're forget about fullName for this option
This argument can't be provided by it's name (not sure, if it's bug or feature). Running `save --testFiles foo` just adds `["--testFiles", "foo"]` to these varargs. Also, normally `fullName` is deducted from variable name, we specify it explicitly, because for some options they are different and we wanted to have all generated options consistent.

petertrr: Actually, adding `testFiles` in JSON generally isn't a very good idea, because it will get added to the options table, and we still will need to tweak logic, so that it's representation in the table is reasonable. Let's keep it hard-coded in generation code for now, I think it's enough to write about these parameters in readme (or in `save --help`)

README.md
```diff
@@ -108,6 +108,17 @@ For example, the following example is invalid and will cause an error, because S
 As described above, `save.toml` is needed to configure tests. The idea is to have only one configuration file for a directory with tests (one to many relation). 
 Such directories we will call `test suites`. We decided to have only one configuration file as we have many times seen that for such tests there is a duplication of configuration in the same test suite.
 
+## Executing specific tests
```
akuleshov7, line 111: Let’s merge it with how to start section (rename it to something like: Usage)

petertrr: I didn't want to add it to `How to start`, because it's rather specific execution mode and I thought we should at first describe basic usage, files hierarchy and only then tell about individual test executions

README.md
```diff
@@ -108,6 +108,17 @@ For example, the following example is invalid and will cause an error, because S
...
+It can be useful to execute only a number of tests instead of all tests under a particular `save.toml` config.
+To do so, you want to pass a test file name:
+```bash
+$ save /path/to/tests/Test1
+```
+or a list of file names (separated with spaces)
+```bash
+$ save /path/to/tests/Test1 /path/to/tests/Test2
```
kgevorkyan, line 119: If for example we will pass some additional arguments, but after `testFiles`
`save /path/to/tests/Test1 /path/to/tests/Test2 -parallel`

Then all of them will be interpret as test files, and there is no workaround, since we use `vararg` and can't use full name for this option?
May be we can at least notice the user, to pass arguments in this case before test files?

petertrr: Good point, added a note about order. Sadly, kotlinx.cli doesn't provide a very good help message for our case:
```bash
$ save --help
Usage: save options_list
Arguments:
    testFiles -> Paths to individual test files, can be provided to execute only them (optional) { String }
Options:
    --test-config, -c -> Path to a configuration of a test suite { String }
   ...
```
I would expect it to state something like `Usage: save options_list [files_list]`, so yes, let's at least state it in the readme

kgevorkyan (approved):

akuleshov7 (approved):

PR #102, opened 2021-05-28
Fix type resolving in IDE in generated code

petertrr:
### What's done:
* Added dependencies for generated source set
* Fixed OptionsTable path in Generation.kt


kgevorkyan (approved):
lgtm

Issue #86, opened 2021-05-25
Add a method for the validation of Config sections in toml configuration


akuleshov7:
We found this on 'tags' section. It is not needed on the high levels of the tree structure. 
We need to validate ONLY config THAT will be used for running tests (and not validate configs from other high levels)

akuleshov7:
Also, under this task you need to remove "required init blocks" in plugin data classes from corresponding data classes.

kgevorkyan:
It also should be provided logic for default values for nullable fields, since in stage of merging we can do this

akuleshov7:
you should validate **ONLY** those configs that contain any test resources

PR #110, opened 2021-05-31
Adding test processing and changing the logic for merging of configs

akuleshov7:
### What's done:
- test processing
- changed merging mechanism

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
@@ -63,7 +63,7 @@ data class TestConfig(
      * @return a [Sequence] of parent config files
      */
     fun parentConfigs(withSelf: Boolean = false) =
-        generateSequence(if (withSelf) this else parentConfig) { it.parentConfig }
+            generateSequence(if (withSelf) this else parentConfig) { it.parentConfig }
 
     /**
      * recursively (till leafs) return all configs from the configuration Tree
@@ -107,15 +107,14 @@ data class TestConfig(
     }
 }
 
+/**
+ * Sections of a toml configuration for tests (including standard plugins)
+ */
+enum class TestConfigSections {
+    FIX, GENERAL, WARN;
+}
 
-    /**
-     * Sections of a toml configuration for tests (including standard plugins)
-     */
-    enum class TestConfigSections {
-        FIX, GENERAL, WARN;
-    }
-
-    /**
-     * @return whether a file denoted by this [Path] is a default save configuration file (save.toml)
-     */
-    fun Path.isSaveTomlConfig() = name == "save.toml"
+/**
+ * @return whether a file denoted by this [Path] is a default save configuration file (save.toml)
+ */
+fun Path.isSaveTomlConfig() = name == "save.toml"
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
@@ -87,9 +87,9 @@ data class TestConfig(
     fun mergeConfigWithParents(): TestConfig {
         logDebug("Start merging configs for ${this.location}")
 
-        this.parentConfigs().toList().reversed().forEach { parentConfig ->
+        this.parentConfigs().toList().forEach { parentConfig ->
             // return from the function if we stay at the root element of the plugin tree
-            val parentalPlugins = parentConfig.pluginConfigs ?: return this
+            val parentalPlugins = parentConfig.pluginConfigs
             parentalPlugins.forEach { currentConfig ->
                 val childConfigs = this.pluginConfigs.filter { it.type == currentConfig.type }
                 if (childConfigs.isEmpty()) {
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
@@ -5,11 +5,11 @@
 package org.cqfn.save.core.config
 
 import org.cqfn.save.core.logging.logDebug
+import org.cqfn.save.core.plugin.GeneralConfig
 import org.cqfn.save.core.plugin.PluginConfig
 
 import okio.FileSystem
 import okio.Path
-import org.cqfn.save.core.plugin.GeneralConfig
 
 /**
  * Configuration for a test suite, that is read from test suite configuration file (toml config)
@@ -63,7 +63,7 @@ data class TestConfig(
      * @return a [Sequence] of parent config files
      */
     fun parentConfigs(withSelf: Boolean = false) =
-        generateSequence(if (withSelf) this else parentConfig) { it.parentConfig }
+            generateSequence(if (withSelf) this else parentConfig) { it.parentConfig }
 
     /**
      * recursively (till leafs) return all configs from the configuration Tree
@@ -81,6 +81,8 @@ data class TestConfig(
 
     /**
      * Merge parent list of plugins with the current list
+     *
+     * @return merged test config
      */
     fun mergeConfigWithParent(): TestConfig {
         logDebug("Start merging configs for ${this.location}")
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
@@ -9,6 +9,7 @@ import org.cqfn.save.core.plugin.PluginConfig
 
 import okio.FileSystem
 import okio.Path
+import org.cqfn.save.core.plugin.GeneralConfig
 
 /**
  * Configuration for a test suite, that is read from test suite configuration file (toml config)
@@ -20,7 +21,7 @@ import okio.Path
 data class TestConfig(
     val location: Path,
     val parentConfig: TestConfig?,
-    val pluginConfigs: MutableList<PluginConfig<*>> = mutableListOf(),
+    val pluginConfigs: MutableList<PluginConfig> = mutableListOf(),
     private val fs: FileSystem = FileSystem.SYSTEM,
 ) {
     /**
@@ -61,7 +62,8 @@ data class TestConfig(
      * @param withSelf if true, include this config as the first element of the sequence or start with parent config otherwise
      * @return a [Sequence] of parent config files
      */
-    fun parentConfigs(withSelf: Boolean = false) = generateSequence(if (withSelf) this else parentConfig) { it.parentConfig }
+    fun parentConfigs(withSelf: Boolean = false) =
+        generateSequence(if (withSelf) this else parentConfig) { it.parentConfig }
 
     /**
      * recursively (till leafs) return all configs from the configuration Tree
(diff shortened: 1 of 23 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -4,27 +4,25 @@
...
     /**
-     * Function which is capable for merging inherited configurations
-     *
-     * @param parentConfig config which will be merged with [this]
-     * @return corresponding merged config
+     * @param otherConfig - 'this' will be merged with 'other'
+     * @return merged config
      */
-    fun mergePluginConfig(parentConfig: T): T
+    fun mergeWith(otherConfig: PluginConfig): PluginConfig
```
petertrr, line 25: But this way we can't guarantee, that `otherConfig` is of the same class. Do you suggest to make checks inside plugins, or will it be like a contract?

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -73,62 +65,73 @@ class Save(
...
-                when (it) {
-                    is FixPluginConfig -> FixPlugin(testConfig)
-                    is WarnPluginConfig -> WarnPlugin(testConfig)
-                    else -> throw PluginException("Unknown type <${it::class}> of plugin config was provided")
-                }
+        ConfigDetector()
+            .configFromFile(fullPathToConfig)
+            .getAllTestConfigs()
+            .forEach { testConfig ->
+                val reporter = getReporter(saveProperties)
```
petertrr, line 72: I think reporter can be created outside of the loop

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -169,7 +172,7 @@ class Save(
         }
     }
 
-    @Suppress("WHEN_WITHOUT_ELSE")  // TestResult is a sealed class
+    @Suppress("WHEN_WITHOUT_ELSE")  // TestStatus is a sealed class
```
petertrr, line 175: why?

save-core/src/commonTest/kotlin/org/cqfn/save/core/PerformAnalysisTest.kt
```diff
@@ -2,10 +2,12 @@ package org.cqfn.save.core
 
 import org.cqfn.save.core.config.ReportType
 import org.cqfn.save.core.config.SaveProperties
+import kotlin.test.Ignore
 import kotlin.test.Test
 
-class PluginDiscoveryTest {
+class PerformAnalysisTest {
     @Test
+    @Ignore
```
petertrr, line 10: I think it should work already, does it crash now?

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -4,27 +4,25 @@
...
+interface PluginConfig {
     /**
-     * Merge current config into [childConfig]
-     *
-     * @param childConfig list of child configs, which will be filtered for merging
+     * type of the config (usually related to the class: WARN/FIX/e.t.c)
      */
-    @Suppress("TYPE_ALIAS")
-    fun mergeConfigInto(childConfig: MutableList<PluginConfig<*>>)
+    val type: TestConfigSections
```
petertrr, line 19: Won't it cause troubles, when we start implementing custom plugins injection? Maybe, instead of enum, we can have a marker interface here, that still could be used to compare plugins, but will be extendible?

petertrr (changes requested):

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -71,50 +73,36 @@ data class TestConfig(
...
 
-    // Merge child configs recursively
-    private fun mergeChildConfigs() {
-        for (child in childConfigs) {
-            child.mergeChildConfigWithParent(parent = this)
-            child.mergeChildConfigs()
+        // return from the function if we stay at the root element of the plugin tree
+        val parentalPlugins = this.parentConfig?.pluginConfigs ?: return this
+        parentalPlugins.forEach { currentConfig ->
+            val childConfigs = this.pluginConfigs.filter { it.type == currentConfig.type }
```
petertrr, line 93: Can we compare `it::class` and `currentConfig::class` in common code?

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -4,27 +4,25 @@
 
 package org.cqfn.save.core.plugin
 
+import org.cqfn.save.core.config.TestConfigSections
+
 import kotlinx.serialization.Serializable
 
 /**
- * Core interface for plugin configuration
+ * Core interface for plugin configuration (like warnPlugin/fixPLuin/e.t.c)
```
kgevorkyan, line 12:
```suggestion
 * Core interface for plugin configuration (like warnPlugin/fixPlugin/e.t.c)
```

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -48,30 +46,24 @@ data class GeneralConfig(
...
+        }
+
         return GeneralConfig(
             mergedTag,
-            this.description ?: parentConfig.description,
-            this.suiteName ?: parentConfig.suiteName,
-            this.excludedTests ?: parentConfig.excludedTests,
-            this.includedTests ?: parentConfig.includedTests,
-            this.ignoreSaveComments ?: parentConfig.ignoreSaveComments
+            this.description,
```
kgevorkyan, line 62: I thought, that we decide to check all fields for nullability in aim of consistency

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -48,30 +46,24 @@ data class GeneralConfig(
...
-    }
+) : PluginConfig {
+    override val type = TestConfigSections.GENERAL
 
-    override fun mergePluginConfig(parentConfig: GeneralConfig): GeneralConfig {
-        val mergedTag = parentConfig.tags?.let {
-            val parentTags = parentConfig.tags.split(", ")
+    override fun mergeWith(otherConfig: PluginConfig): PluginConfig {
+        val other = otherConfig as GeneralConfig
+        val mergedTag = other.tags.let {
```
kgevorkyan, line 54: This logic was unclear before too, since in current version this field is not nullable. However, in https://github.com/cqfn/save/issues/86 I will make it nullable. So, the proper code will be some thing like this
```kotlin
val mergedTag = other.tags?.let {
    this.tags?.let {
        val parentTags = other.tags.split(", ")
        val childTags = this.tags.split(", ")
        parentTags.union(childTags).joinToString(", ")
    } ?: other.tags
} ?: this.tags
``` 
I think you can refuse of this `let` at all, since your fields aren't nullable, and I will change it in PR of validation

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -73,62 +65,73 @@ class Save(
...
-            .pluginConfigs.filterNot { it is GeneralConfig }
-            .map {
-                when (it) {
-                    is FixPluginConfig -> FixPlugin(testConfig)
-                    is WarnPluginConfig -> WarnPlugin(testConfig)
-                    else -> throw PluginException("Unknown type <${it::class}> of plugin config was provided")
-                }
+        ConfigDetector()
+            .configFromFile(fullPathToConfig)
+            .getAllTestConfigs()
```
kgevorkyan, line 70: `getAllTestConfigs()` returns list of `this` and child configs. And you merge them pairwise. However, previously, merge function merged all parents configs for current, and not only the first parent.

What will do your solution, in case if we will give some config in the middle of the hierarchy tree?

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7: I thouht about pluses and minuses of that. Pluses:
- simple code 
- composition design pattern used

Minuses:
- we can face exceptions in Plugin code, **but** it will be easily caught on the first run as this method is called from single place where we have an explicit type check. **What could have gone wrong?** 👯‍♂️

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
akuleshov7: oh, sure

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
akuleshov7: TestResult is not a sealed class )
TestStatus is

save-core/src/commonTest/kotlin/org/cqfn/save/core/PerformAnalysisTest.kt
akuleshov7: yeah, because oh not implemented reporter

petertrr: I've added a simple reporter in #113, later we can use some nice library to properly format a table

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7: yes, very good point I see the problem, but let's leave it as is by for now... 
Anyway it will be a hevay change

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
akuleshov7: YES! And also I forot that in getAllConfigs flat map is not sorted by levels

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7: fixed for static analysis in idea

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7: fixed

save-core/src/commonTest/kotlin/org/cqfn/save/core/PerformAnalysisTest.kt
akuleshov7: removed ignore

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
akuleshov7: discussed - these are problems of future us

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7: fixed to your version

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
akuleshov7: fixed

save-core/src/commonTest/kotlin/org/cqfn/save/core/MergeConfigsTest.kt
```diff
@@ -141,10 +141,10 @@ class MergeConfigsTest {
         val config3 = TestConfig(toml3, config2, mutableListOf(generalConfig3, warnConfig3, fixConfig3))
         val config4 = TestConfig(toml4, config3, mutableListOf(generalConfig4, warnConfig4, fixConfig4))
 
-        config1.mergeConfigWithParent()
-        config2.mergeConfigWithParent()
-        config3.mergeConfigWithParent()
-        config4.mergeConfigWithParent()
+        config1.mergeConfigWithParents()
```
kgevorkyan, line 144: I think we don't need merge each config with it parent. The main logic of such tests - call the merge function from some place in hierarchy tree (top, bottom, middle) and check, that all configs were successfully merged

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
kgevorkyan: Seems, like it still `fixPluin` and `e.t.c`

save-core/src/commonTest/kotlin/org/cqfn/save/core/MergeConfigsTest.kt
```diff
@@ -169,9 +166,6 @@ class MergeConfigsTest {
         val config3 = TestConfig(toml3, config2, mutableListOf(generalConfig3, warnConfig3, fixConfig3))
         val config4 = TestConfig(toml4, config3, mutableListOf(generalConfig4))
 
-        config1.mergeConfigWithParents()
-        config2.mergeConfigWithParents()
-        config3.mergeConfigWithParents()
         config4.mergeConfigWithParents()
```
kgevorkyan, line 169: Here we suppose to start from the top config, i.e. from `config1`

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -71,61 +73,49 @@ data class TestConfig(
...
+     * Merge parent list of plugins with the current list
+     *
+     * @return merged test config
+     */
+    fun mergeConfigWithParents(): TestConfig {
+        logDebug("Start merging configs for ${this.location}")
 
-        pairs.forEach { (parent, child) ->
-            child.mergeChildConfigWithParent(parent)
+        this.parentConfigs().toList().forEach { parentConfig ->
```
petertrr, line 90: Is it the correct logic? We should start merging from the root config.
Also, if we are starting from the middle, then we should traverse parents once, and then go down from the provided node

save-core/src/commonTest/kotlin/org/cqfn/save/core/MergeConfigsTest.kt
akuleshov7: yes

petertrr (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
akuleshov7: discussed

PR #112, opened 2021-06-01
Split execCmd to two options

### LINKED ISSUE #107: Split execCmd to two options: one in general config and second in the plugin
What to do when we need to configure a path to a tool that we are testing for the whole project?
It will be user-friendly to inherit and have a common mechanism for a configuration of "execCmd".

So "execCmd" can be moved to general section. **BUT** you need also to support "execFlags" in corresponding plugins.

Cheshiriks:
### What's done:
* Split execCmd to two options: one in general config and second in the plugin

Closes #107


examples/highlevel/suite1/save.toml
```diff
@@ -1,10 +1,11 @@
 [general]
+execCmd = ""
 tags = tag1
 description = "My suite description"
 suiteName = "DocsCheck"
 
 [warn]
-execCmd = "./detekt --build-upon-default-config -i"
+execFlags = "./detekt --build-upon-default-config -i"
```
petertrr, line 8: Same here

examples/highlevel/suite1/save.toml
```diff
@@ -14,7 +15,7 @@ warningTextHasColumn = true
 warningTextHasLine = true
 
 [diff]
-exec_cmd="./ktlint -R diktat-0.4.2.jar -F"
+exec_flags="./ktlint -R diktat-0.4.2.jar -F"
```
petertrr, line 18: This was an invalid example, we can't use different binaries in the same suite. Please remove ktlint from here; to keep example complete, use autofix mode of detekt

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -36,11 +37,13 @@ class FixPlugin(testConfig: TestConfig, testFiles: List<String> = emptyList()) :
 
     override fun handleFiles(files: Sequence<List<Path>>): Sequence<TestResult> {
         val fixPluginConfig = testConfig.pluginConfigs.filterIsInstance<FixPluginConfig>().single()
+        val generalConfig = testConfig.pluginConfigs.filterIsInstance<GeneralConfig>().singleOrNull()
         logInfo("Discovered the following file pairs for comparison: $files")
         return files
             .map { it.first() to it.last() }
             .map { (expected, test) ->
-                val executionResult = pb.exec(fixPluginConfig.execCmd, null, false)
+                val execCmd = generalConfig?.execCmd + " " + fixPluginConfig.execFlags
```
petertrr, line 45: What should we do, if general config is null? I think, we should either raise a warning or use `execFlags` only instead of `execCmd`

save-plugins/warn-plugin/README.md
```diff
@@ -59,7 +59,7 @@ description = "My suite description"
 suiteName = "DocsCheck"
 
 [warn]
-execCmd = "./detekt --build-upon-default-config -i"
+execFlags = "./detekt --build-upon-default-config -i"
```
petertrr, line 62: Please change this exampe

save-plugins/warn-plugin/README.md
```diff
@@ -70,7 +70,7 @@ warningTextHasLine = true
 ```
 
 When executed from project root (where `save.propertes` is located), SAVE will cd to `rootDir` and discover all files
-matching `inputFilePattern`. It will then execute `$exec_cmd $testFile` (since we specified
+matching `inputFilePattern`. It will then execute `$exec_flags $testFile` (since we specified
```
petertrr, line 73:
```suggestion
matching `inputFilePattern`. It will then execute `$exec_cmd $exec_flags $testFile` (since we specified
```

petertrr (changes requested):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -73,9 +73,9 @@ class WarnPlugin(testConfig: TestConfig, testFiles: List<String> = emptyList())
 
         val execCmd: String = if (generalConfig?.ignoreSaveComments == true) {
             val fileName = createTestFile(path, warnPluginConfig.warningsInputPattern)
-            warnPluginConfig.execCmd + " $fileName"
+            generalConfig.execCmd + " " + warnPluginConfig.execFlags + " $fileName"
```
akuleshov7, line 76: please make a single string

akuleshov7 (approved):

akuleshov7 (approved):

petertrr (approved):

akuleshov7 (approved):

PR #116, opened 2021-06-02
Add a method for the validation of config sections in toml configuration

kgevorkyan:
### What's done:
* Provide the methods for plugin configs validation
* Add tests, update README
* Add new reporter for plugin skipping


save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -105,6 +105,16 @@ data class TestConfig(
         }
         return this
     }
+
+    /**
+     * Method, which validates all plugin configs and set default values, if possible
+     */
+    fun validateAndSetDefaults() {
+        logDebug("Start plugin validation for $location")
+        for (config in pluginConfigs) {
```
petertrr, line 114: `forEachIndexed` instead of calling `indexOf` on every iteration?

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -62,13 +69,40 @@ data class GeneralConfig(
...
+            tags,
+            description,
+            suiteName,
+            excludedTests ?: "",
+            includedTests ?: "",
+            ignoreSaveComments ?: false
+        )
+    }
+
+    private fun errorMsgForRequireCheck(field: String) = "Error: Couldn't found `$field` in [general] section. Please provide it in this, " +
```
petertrr, line 106:
```suggestion
    private fun errorMsgForRequireCheck(field: String) = "Error: Couldn't find `$field` in [general] section. Please provide it in this, " +
```
Also, can we specify, which config is `this` here?

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -92,14 +92,19 @@ class Save(
         logInfo("=> Executing plugin: ${plugin::class.simpleName} for [${plugin.testConfig.location}]")
         reporter.onPluginExecutionStart(plugin)
         try {
-            plugin.execute()
+            val events = plugin.execute()
+            if (events.toList().isEmpty()) {
+                logWarn("No resources discovered for ${plugin::class.simpleName} in [${plugin.testConfig.location}], skipping")
```
petertrr, line 97: Why is it warn? I think, we consider this normal behavior, so it should be debug?

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -38,14 +37,16 @@ class FixPlugin(testConfig: TestConfig, testFiles: List<String> = emptyList()) :
         .build()
 
     override fun handleFiles(files: Sequence<List<Path>>): Sequence<TestResult> {
-        val fixPluginConfig = testConfig.pluginConfigs.filterIsInstance<FixPluginConfig>().single()
-        val generalConfig = testConfig.pluginConfigs.filterIsInstance<GeneralConfig>().singleOrNull()
         val flattenedResources = files.toList()
         if (flattenedResources.isEmpty()) {
```
petertrr, line 41:
```suggestion
        if (files.any { it.isNotEmpty() }) {
```
we don't need the whole sequence here

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -33,15 +32,16 @@ class WarnPlugin(testConfig: TestConfig, testFiles: List<String> = emptyList())
     override fun handleFiles(files: Sequence<List<Path>>): Sequence<TestResult> {
         val flattenedResources = files.toList().flatten()
         if (flattenedResources.isEmpty()) {
```
petertrr, line 34:
```suggestion
        if (files.any { it.isNotEmpty() }) {
```

petertrr (changes requested):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7, line 106: I suppose that if it is easy - we should add a location of a test config with a problem

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -92,14 +92,19 @@ class Save(
...
+            }
+            events
                 .onEach { event -> reporter.onEvent(event) }
                 .forEach(this::handleResult)
         } catch (ex: PluginException) {
             reporter.onPluginExecutionError(ex)
             logError("${plugin::class.simpleName} has crashed: ${ex.message}")
         }
-        logInfo("<= Executed plugin: ${plugin::class.simpleName} for [${plugin.testConfig.location}]")
+        logInfo("<= Finish execution of: ${plugin::class.simpleName} for [${plugin.testConfig.location}]")
```
akuleshov7, line 107: finished

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
```diff
@@ -31,41 +31,72 @@ import kotlinx.serialization.UseSerializers
...
+            this.warningsInputPattern ?: other.warningsInputPattern,
+            this.warningsOutputPattern ?: other.warningsOutputPattern,
             this.warningTextHasLine ?: other.warningTextHasLine,
             this.warningTextHasColumn ?: other.warningTextHasColumn,
             this.lineCaptureGroup ?: other.lineCaptureGroup,
             this.columnCaptureGroup ?: other.columnCaptureGroup,
-            this.messageCaptureGroup,
+            this.messageCaptureGroup ?: other.columnCaptureGroup,
             this.exactWarningsMatch ?: other.exactWarningsMatch,
             this.testNameSuffix ?: other.testNameSuffix
```
akuleshov7, line 65: don't see a default for testNameSuffix

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
```diff
@@ -20,7 +20,7 @@ import kotlinx.serialization.UseSerializers
  */
 @Serializable
 data class FixPluginConfig(
-    val execFlags: String,
+    val execFlags: String? = null,
     val resourceNameTestSuffix: String? = null,
```
akuleshov7, line 24: where are defaults for suffixes?

akuleshov7 (changes requested):
some defaults should be confirmed

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
akuleshov7: yeah, this is my code, may be should be changed to INFO, if user has forgotten to put resources

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
kgevorkyan: Yes, it will be more appropriate, fixed

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
kgevorkyan: Fixed to `find`.
About specification - I was thinking about specifying the location, but in PluginConfig we have no such information, only in TestConfig.
I added at least printing of the fields of current config.

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
kgevorkyan: Changed to INFO

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
kgevorkyan: Agree, for `if` we don't need the whole sequence, however we use it for `logInfo`

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
kgevorkyan: Please take a look at https://github.com/cqfn/save/pull/116#discussion_r645447932

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
kgevorkyan: Thanks, fixed

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
kgevorkyan: We have separate variables `testName` for `warn` `resourceNameTest` `resourceNameExpected` which are already who perform this role, and I use them as arguments in validation

The reason, why we still need such variables - we use them for resource detection, which is doing before validation

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
kgevorkyan: Same explanation as in https://github.com/cqfn/save/pull/116#discussion_r645449191

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7: please create an issue to add TestConfig field to the PluginConfig to support this kind of logging

akuleshov7 (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
kgevorkyan: Created https://github.com/cqfn/save/issues/118

petertrr (approved):

PR #120, opened 2021-06-04
Handle execution output from ProcessBuilder in Plugins

kgevorkyan:
### What's done:
* Handle execution output from ProcessBuilder in Plugins
* Update toml example files
* Fix in ProcessBuilder native: if command is complex, all output will be redirected into one variable (fixed by adding '(' ')' in cmd)
* Change ERRORLEVEL for echo commands on Windows from 1 to 0
* Update tests

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -11,12 +11,12 @@ import okio.Path
 /**
  * Plugin that can be injected into SAVE during execution. Plugins accept contents of configuration file and then perform some work.
  * @property testConfig
- * @property collectStdOut
+ * @property useInternalRedirections whether to redirect stdout/stderr for internal purposes
  */
 abstract class Plugin(
     open val testConfig: TestConfig,
     private val testFiles: List<String>,
-    protected val collectStdOut: Boolean) {
+    protected val useInternalRedirections: Boolean) {
     private val fs = FileSystem.SYSTEM
 
     /**
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -11,8 +11,12 @@ import okio.Path
 /**
  * Plugin that can be injected into SAVE during execution. Plugins accept contents of configuration file and then perform some work.
  * @property testConfig
+ * @property collectStdOut
  */
-abstract class Plugin(open val testConfig: TestConfig, private val testFiles: List<String>) {
+abstract class Plugin(
+    open val testConfig: TestConfig,
+    private val testFiles: List<String>,
+    protected val collectStdOut: Boolean) {
     private val fs = FileSystem.SYSTEM
 
     /**
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -4,6 +4,7 @@ import org.cqfn.save.core.config.TestConfig
 import org.cqfn.save.core.config.isSaveTomlConfig
 import org.cqfn.save.core.files.findDescendantDirectoriesBy
 import org.cqfn.save.core.result.TestResult
+import org.cqfn.save.core.utils.ProcessBuilder
 
 import okio.FileSystem
 import okio.Path
@@ -16,8 +17,13 @@ import okio.Path
 abstract class Plugin(
     open val testConfig: TestConfig,
     private val testFiles: List<String>,
-    protected val useInternalRedirections: Boolean) {
+    private val useInternalRedirections: Boolean) {
     private val fs = FileSystem.SYSTEM
+
+    /**
+     * Instance that is capable of executing processes
+     */
+    val pb = ProcessBuilder(useInternalRedirections)
 
     /**
      * Perform plugin's work.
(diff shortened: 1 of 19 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -11,8 +11,12 @@ import okio.Path
 /**
  * Plugin that can be injected into SAVE during execution. Plugins accept contents of configuration file and then perform some work.
  * @property testConfig
+ * @property useInternalRedirections whether to redirect stdout/stderr for internal purposes
  */
-abstract class Plugin(open val testConfig: TestConfig, private val testFiles: List<String>) {
+abstract class Plugin(
+    open val testConfig: TestConfig,
+    private val testFiles: List<String>,
+    protected val useInternalRedirections: Boolean) {
```
petertrr, line 19: In which case would we need to change this parameter? Does it make sense to make it configurable from plugins, when extending this class, or can it be configured globally application-wise?

kgevorkyan: We need to run tests with `useInternalRedirections = false`, since we use `echo >` in them.
In other cases we use `useInternalRedirections = true`, for now. 

Previously it was hard coded for FixPlugin to `false`, which is incorrect, since we need to collect `stderr` and `stdout` in the course of work, so I decide to make it configurable.

In general, this argument just a workaround, since we use redirections for two purposes:
1) collect process output, by creating temp files
2) creating new files with some info (like we do in tests)
and unfortunately we can't do this at the same time. Only the last redirection in bash command will be applied.

So, the configuration must be hidden or carried out with full awareness of the consequences

save-common/src/nativeMain/kotlin/org/cqfn/save/core/utils/ProcessBuilderInternal.kt
```diff
@@ -12,12 +12,12 @@ import platform.posix.system
 actual class ProcessBuilderInternal actual constructor(
     private val stdoutFile: Path,
     private val stderrFile: Path,
-    private val collectStdout: Boolean
+    private val useInternalRedirections: Boolean
 ) {
-    actual fun prepareCmd(command: String) = if (collectStdout) {
-        "$command >$stdoutFile 2>$stderrFile"
+    actual fun prepareCmd(command: String) = if (useInternalRedirections) {
+        "($command) >$stdoutFile 2>$stderrFile"
```
kgevorkyan, line 18: We need brackets here, since if command is complex (contain `&&` or `;`) we need to redirect output from all command at the one time.

The minus of this approach - now we can't redirect output, if command already have redirections.
`(cmd > /dev/null)` > `>$stdoutFile 2>$stderrFile` - redirection to `/dev/null` will applied at any time.
However, we don't allow user redirections, so I think it's okay. We use redirections in cmd only for tests

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
petertrr: I'd prefer to hide it from plugin implementations, however it seems tricky to do. We could make it `private` in `Plugin` and also create `useInternalRedirections` in `ProcessBuilder` itself (because we won't need to change it between invocations of the same ProcessBuilder, if I understand you correctly). Then we can move `val pb = ProcessBuilder(useInternalRedirections)` to the base `Plugin` class, so the only place, where plugin implementation sees this parameter is constructor invocation. What do you think?

kgevorkyan: Yes, thanks, it will be even better. Changed

petertrr (approved):
LGTM

save-plugins/warn-plugin/README.md
```diff
@@ -83,7 +83,7 @@ testNameSuffix = "Test" # (default value)
 ```
 
 When executed from project root (where `save.propertes` is located), SAVE will cd to `rootDir` and discover all files
```
akuleshov7, line 85: please write to main readme: "Your application should return 0 in case of successful processing, otherwise save will not be able to determine correct result"

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -98,11 +102,14 @@ class WarnPlugin(testConfig: TestConfig, testFiles: List<String> = emptyList())
...
-            if (generalConfig?.ignoreSaveComments == true) createTestFile(it, warnPluginConfig.warningsInputPattern!!) else it.name
+            if (generalConfig!!.ignoreSaveComments == true) createTestFile(it, warnPluginConfig.warningsInputPattern!!) else it.toString()
         }
-        val execCmd = "${(generalConfig?.execCmd ?: "")} ${warnPluginConfig.execFlags} $fileNames"
+        val execCmd = "${generalConfig!!.execCmd} ${warnPluginConfig.execFlags} $fileNames"
 
         val executionResult = pb.exec(execCmd, null)
+        val stdout = executionResult.stdout.joinToString("\n")
+        val stderr = executionResult.stderr.joinToString("\n")
+        val status = if (executionResult.code != 0) Fail(stderr) else null
```
akuleshov7, line 112: else checkResult()

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -137,7 +144,7 @@ class WarnPlugin(testConfig: TestConfig, testFiles: List<String> = emptyList())
                 }
             }
         }
-        return fileName
+        return fileName.toString()
```
akuleshov7, line 147: you can remove toString here

akuleshov7 (approved):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
kgevorkyan: Thanks, fixed

save-plugins/warn-plugin/README.md
kgevorkyan: Done

PR #123, opened 2021-06-08
Provide field of config location for PluginConfig

kgevorkyan:
### What's done:
* Provide field of config location for PluginConfig

### DIFF
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
@@ -7,7 +7,11 @@ import org.cqfn.save.core.config.TestConfigSections
 import org.cqfn.save.core.plugin.PluginConfig
 import org.cqfn.save.core.utils.RegexSerializer
 
+import okio.Path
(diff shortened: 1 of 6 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
```diff
@@ -57,6 +61,9 @@ data class WarnPluginConfig(
     val testNameSuffix: String? = null,
 ) : PluginConfig {
     override val type = TestConfigSections.WARN
+
+    @Transient
+    override var configLocation: Path = "undefined_toml_location".toPath()
```
akuleshov7, line 66: please check how ktoml will handle it

kgevorkyan: I add `configLocation` to toml file and received the following:
`Unknown key received: <configLocation> in scope ...`, so I think all is okay, and `@Transient` hides this variable from the
serialization process.

I also enabled `PerformAnalysisTest`, since after https://github.com/cqfn/save/pull/120 all worked fine.

akuleshov7 (approved):

Issue #126, opened 2021-06-09
fix-plugin: retain file name and file path when creating a tmp copy


petertrr:
because the tool under testing may rely on these data for some checks. Probably, we should create a copy of all resources

petertrr:
Also important for warn-plugin, see #135

Issue #125, opened 2021-06-09
Use save.properties from pwd as default value


petertrr:
When `save-cli` is built from a9362a66, and directory structure is as follows:
```
$ ls
save save.properties save.toml tests
```
and save is executed as `./save`, it doesn't read `save.proeprties`, but should, according to the default value

kgevorkyan:
Previously we didn't consider default values for config options at all in course of work of SAVE, because
in SaveProperties was the following logic:
```kotlin
parser.parse(args)
this.testConfigPath = testConfigPath
this.parallelMode = parallelMode
...
```
and this logic overwrites all default values by `null` if they weren't provided in CLI, this is how parser works.
Default values worked only in `performaAnalysisTest`, since we created instance manually without nulls.

We also can't change this logic to something like this (overwrite values only if corresponding field from CLI != `null`)
```kotlin
testConfigPath?.let { this.testConfigPath = testConfigPath }
```
because in this case we won't be able to identify in `mergeConfigWithPriorityToThis()` whether the value was passed by user, or taken from default (situation like for default values in plugin configs)

It was decided to make all fields in SaveProperties `null` by default and to create some instance with explicit values.
So process of merging will be as following:
```kotlin
defaultSavePropertiesInstanse = SaveProperties(testConfigPath = "save.toml", parallelMode = false, ....)
....
....
// Merging:
testConfigPath = testConfigPath ?: configFromPropertiesFile.testConfigPath ?: defaultSavePropertiesInstanse.testConfigPath
```

PR #128, opened 2021-06-09
Create example with diktat for Kotlin

### LINKED ISSUE #99: Create an external repository for test examples
Let's test clang: -> llvm ir -> execute binary

petertrr:
### What's done:
* Added configs and test resources for `examples/kotlin-diktat`
* Added run.sh for setting up environment for this example
* Moved old examples into `examples/test-discovery`

Closes #99 

Found the following bugs so far: 
* Critical: ~#130 (PR)~, ~#135~
* Not very critical: #125, #126, #131, #133, #134, #137

### TODO:
* [ ] Ktoml doesn't support quotes yet, for now our configs are with plain text


examples/kotlin-diktat/warn/save.toml
```diff
@@ -0,0 +1,20 @@
+[general]
+tags = warn
+description = "Test warnings discovered by diKTat"
+suiteName = "warnings"
+ignoreSaveComments = true
+
+[warn]
+execFlags = --debug  # todo: replace with empty string once ktoml supports quotes
```
kgevorkyan, line 8: Is it just placeholder, while you're writing, that we need to replace it with empty string in future?
May be we can just just don't specify `execFlags` at all? Validation method should change `null` to the empty string.

examples/kotlin-diktat/fix/smoke/save.toml
```diff
@@ -0,0 +1,6 @@
+[general]
+tags = "smoke"
```
kgevorkyan, line 2:
```suggestion
tags = smoke
```
For now?

examples/kotlin-diktat/run.sh
```diff
@@ -0,0 +1,19 @@
...
+wget -nc https://github.com/cqfn/diKTat/releases/download/v0.6.2/diktat-0.6.2.jar
+# Download latest release of SAVE CLI
+#wget -nc https://github.com/cqfn/save/releases/download/v0.1.0-alpha.2/save-0.1.0-alpha.2-linuxX64.kexe -O save
+# use this for mingw
+#wget -nc https://github.com/cqfn/save/releases/download/v0.1.0-alpha.2/save-0.1.0-alpha.2-mingwX64.exe -O save
+# ... or copy locally built save-cli
+# cp ../../save-cli/build/bin/linuxX64/releaseExecutable/save-0.1.0-alpha.2.66+20210609T095346Z-linuxX64.exe save
+if [ ! -f save ]; then echo "ERROR: Please place save executable in $PWD before continuing" && exit 1; fi
+chmod +x save
+# Run all tests
```
kgevorkyan, line 13: There are a lot of commented out lines, may be we can pass some arguments to the bash script, and use `if-else` statements?
Something like this
```bash
if [ "$1" == "someValue" ]
then
  # Run corresponding test
elif [ ... ]
then
  # Run another test
else
  # Run something else
fi
```
To avoid running script without args you can do something like this:
```bash
if [[ $# -eq 0 ]] ; then
    echo 'some message'
    exit 1
fi
```

examples/kotlin-diktat/run.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+#wget -nc https://github.com/pinterest/ktlint/releases/download/0.39.0/ktlint
+wget -nc https://repo1.maven.org/maven2/com/pinterest/ktlint/0.39.0/ktlint-0.39.0.jar
+wget -nc https://github.com/cqfn/diKTat/releases/download/v0.6.2/diktat-0.6.2.jar
+# Download latest release of SAVE CLI
+#wget -nc https://github.com/cqfn/save/releases/download/v0.1.0-alpha.2/save-0.1.0-alpha.2-linuxX64.kexe -O save
+# use this for mingw
```
kgevorkyan, line 7: Probably we can use `"$OSTYPE"`
https://stackoverflow.com/a/8597411

examples/kotlin-diktat/warn/save.toml
petertrr: I originally added it because we got an NPE if no `execFlags` have been specified. I believe we have fixed it already, so I'll remove it from here.

examples/kotlin-diktat/fix/smoke/save.toml
petertrr: Yeah, thanks, I'll remove all quotes for consistency.

examples/kotlin-diktat/run.sh
petertrr: Refactored the script

examples/kotlin-diktat/run.sh
petertrr: Thx, added

kgevorkyan (approved):
lgtm

PR #136, opened 2021-06-11
Fixes in warn-plugin

### LINKED ISSUE #135: warn-plugin: Problems with warn-plugin
* [ ] Copy file with same name (Path can be covered in #126, but name is critical)
* [ ] Problems with warnings comparison (there is the smae warning in both expected and actual, but we don't see this):
  ```
  Some warnings were expected but not received:
  [[Warning(message=[ENUM_VALUE] enum values should be in selected UPPER_CASE snake/PascalCase format: NAme_MYa_sayR_, line=8, column=5, fileName=EnumValueSnakeCaseTest.kt)]],
  and others were unexpected:
  [[Warning(message=[ENUMS_SEPARATED] enum is incorrectly formatted: enums must end with semicolon, line=8, column=5, fileName=EnumValueSnakeCaseTest.kt),
  Warning(message=[ENUMS_SEPARATED] enum is incorrectly formatted: last enum entry must end with a comma, line=8, column=5, fileName=EnumValueSnakeCaseTest.kt),
  Warning(message=[ENUM_VALUE] enum values should be in selected UPPER_CASE snake/PascalCase format: NAme_MYa_sayR_, line=8, column=5, fileName=EnumValueSnakeCaseTest.kt)]
  ```

petertrr:
### What's done:
* warn-plugin: retain file name when creating a copy without technical comments
* warn-plugin: fixed logic of warnings comparison
* Added tests

Closes #135

### DIFF
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
@@ -168,10 +168,8 @@ class WarnPlugin(
     internal fun checkResults(expectedWarningsMap: Map<LineColumn?, List<Warning>>,
                               actualWarningsMap: Map<LineColumn?, List<Warning>>,
                               warnPluginConfig: WarnPluginConfig): TestStatus {
-        val missingWarnings: List<Warning> = expectedWarningsMap.values.flatten()
-            .filter { it !in actualWarningsMap.values.flatten() }
-        val unexpectedWarnings: List<Warning> = actualWarningsMap.values.flatten()
-            .filter { it !in expectedWarningsMap.values.flatten() }
+        val missingWarnings = expectedWarningsMap.valuesNotIn(actualWarningsMap)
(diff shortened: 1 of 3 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -176,11 +186,4 @@ class WarnPlugin(
             else -> Fail("")
         }
     }
-
-    private fun testFileName(): String = "test_file${atomicInt.addAndGet(1)}"
```
kgevorkyan, line 180: I thought, that such unique name was created in aim to avoid overwriting in multi thread execution.
What we decide to do now?

petertrr: Yes, but turns out we need to preserve the original file name. We also have an issue to preserve full path (#126), so after that these copied files will be unique automatically.

kgevorkyan (approved):

PR #138, opened 2021-06-18
In batchMode a list of TestResults should be returned instead of a single aggregated TestResult

### LINKED ISSUE #131: warn-plugin: In batchMode a list of TestResults should be returned instead of a single aggregated TestResult
`WaqrnPlguin.handleTestFile` should compare warnings per file and then return a collection of TestResults

Cheshiriks:
What's done:
* WaqrnPlguin.handleTestFile return a collection of TestResults
Closes #131

### DIFF
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
@@ -48,7 +48,7 @@ class WarnPlugin(
 
         return files.chunked(warnPluginConfig.batchSize ?: 1).map { chunk ->
             handleTestFile(chunk.map { it.single() }, warnPluginConfig, generalConfig)
-        }.flatMap { it }
+        }.flatten()
     }
 
     override fun rawDiscoverTestFiles(resourceDirectories: Sequence<Path>): Sequence<List<Path>> {
@@ -116,8 +116,8 @@ class WarnPlugin(
                 DebugInfo(null, ex.message, null)
             ))
         }
-        val stdout = executionResult.stdout.joinToString("\n")
-        val stderr = executionResult.stderr.joinToString("\n")
+        val stdout = executionResult.stdout
+        val stderr = executionResult.stderr
 
         val actualWarningsMap = executionResult.stdout.mapNotNull {
             with(warnPluginConfig) {
@@ -135,7 +135,10 @@ class WarnPlugin(
                     actualWarningsMap.filter { it.value.any { warning -> warning.fileName == path.name } },
                     warnPluginConfig
                 ),
-                DebugInfo(stdout, stderr, null)
+                DebugInfo(
+                    stdout.filter { it.contains(path.name) }.joinToString("\n"),
+                    stderr.filter { it.contains(path.name) }.joinToString("\n"),
+                    null)
             )
         }
     }
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
@@ -34,7 +34,6 @@ class WarnPlugin(
     useInternalRedirections) {
     private val fs = FileSystem.SYSTEM
 
-    @Suppress("LOCAL_VARIABLE_EARLY_DECLARATION")
     override fun handleFiles(files: Sequence<List<Path>>): Sequence<TestResult> {
         val flattenedResources = files.toList().flatten()
         if (flattenedResources.isEmpty()) {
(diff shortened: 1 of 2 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -126,11 +132,22 @@ class WarnPlugin(
             .groupBy { if (it.line != null && it.column != null) it.line to it.column else null }
             .mapValues { (_, warning) -> warning.sortedBy { it.message } }
 
-        return TestResult(
-            paths,
-            checkResults(expectedWarnings, actualWarningsMap, warnPluginConfig),
-            DebugInfo(stdout, stderr, null)
-        )
+        val testResultList: MutableList<TestResult> = mutableListOf()
+        paths.forEach { path ->
```
akuleshov7, line 136: why do you need forEach here with mutable list, use map instead in single functional manner

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -46,9 +47,12 @@ class WarnPlugin(
         val warnPluginConfig = testConfig.pluginConfigs.filterIsInstance<WarnPluginConfig>().single()
         val generalConfig = testConfig.pluginConfigs.filterIsInstance<GeneralConfig>().singleOrNull()
 
-        return files.chunked(warnPluginConfig.batchSize ?: 1).map { chunk ->
+        val listTestResult: MutableList<TestResult> = mutableListOf()
+        files.chunked(warnPluginConfig.batchSize ?: 1).map { chunk ->
             handleTestFile(chunk.map { it.single() }, warnPluginConfig, generalConfig)
-        }
+        }.forEach { listTestResult.addAll(it) }
```
akuleshov7, line 53: same, for each looks strange when you can simply map

akuleshov7 (changes requested):
do it in a more functional way

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -46,9 +47,12 @@ class WarnPlugin(
         val warnPluginConfig = testConfig.pluginConfigs.filterIsInstance<WarnPluginConfig>().single()
         val generalConfig = testConfig.pluginConfigs.filterIsInstance<GeneralConfig>().singleOrNull()
 
-        return files.chunked(warnPluginConfig.batchSize ?: 1).map { chunk ->
+        val listTestResult: MutableList<TestResult> = mutableListOf()
```
petertrr, line 50: If diktat complains on early declaration here, than it's a bug in diktat, because we should be correctly detecting such cases. Please check if it is reproducible and open an issue in that case

Cheshiriks: https://github.com/cqfn/diKTat/issues/934

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -48,7 +48,7 @@ class WarnPlugin(
 
         return files.chunked(warnPluginConfig.batchSize ?: 1).map { chunk ->
             handleTestFile(chunk.map { it.single() }, warnPluginConfig, generalConfig)
-        }
+        }.flatMap { it }
```
petertrr, line 51:
```suggestion
        }
        .flatten()
```

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -126,11 +127,17 @@ class WarnPlugin(
...
-        )
+        return paths.map { path ->
+            TestResult(
+                listOf(path),
+                checkResults(
+                    expectedWarnings.filter { it.value.any { warning -> warning.fileName == path.name } },
+                    actualWarningsMap.filter { it.value.any { warning -> warning.fileName == path.name } },
+                    warnPluginConfig
+                ),
+                DebugInfo(stdout, stderr, null)
```
petertrr, line 138: stdout and stderr are still combined, right? We need to think how to remove redundant copies of this data

petertrr (changes requested):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
Cheshiriks: fixed

petertrr (approved):

akuleshov7 (approved):

PR #139, opened 2021-06-18
API for plugins and plugin configs access

### LINKED ISSUE #108: Add public APIs for integration with save-cloud
Will be used from save-cloud to discover test suites names (from general config) and other info

* API to access list of discovered plugins (public method in `Save.kt`?)
* Can `testConfigPath` and `testRootPath` from `SaveProperties` be merged? Now the only place where we use them is to get full path to config, which we can store in properties right away.
* ~Methods to traverse config hierarchy, computing some values for each config (like map, flatmap etc.). Probably, should let traverse only the configs with non-empty test resources.~ Are not needed now.

Specific place in save-cloud, where this will be used, is `TestDiscoveringService` in `save-preprocessor`

petertrr:
### What's done:
* Moved logic into TestConfig
* Added TestConfigUtils in save-core

Closes #108 

Complementary PR in save-cloud: https://github.com/cqfn/save-cloud/pull/139. The logic we need already can be implemented there, but with a lot of code duplication. This PR proposes to encapsulate it in SAVE core.

Some drawabacks:
*  We can't create specific plugins/pluginConfigs in save-common because of circular dependencies. I've split those methods, so that specific implementation an be kept in save-core, while main logic in save-common
* ~The first WIP variant splits configs loop into two: first fully build configs, then execute. Do we need to keep it as it was? Or do we need to add `Plugin`s in config too?~ Fixed; but point about plugin inclusion still stands, maybe for future PRs.


save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -74,6 +75,38 @@ data class TestConfig(
...
     }
 
+    /**
+     * Walk all descendant configs and merge them with their parents
+     *
+     * @param createPluginConfigList a function which can create a list of [PluginConfig]s for this [TestConfig]
+     * @return an update this [TestConfig]
+     */
+    fun withEvaluatedDescendantConfigs(createPluginConfigList: (TestConfig) -> List<PluginConfig>): TestConfig {
+        getAllTestConfigs().forEach { testConfig ->
```
kgevorkyan, line 85: Do we need `getAllTestConfigs` here for save-cloud? And can't move this function into `forEach` from Save.kt?
Now we goes through all configs twice

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TestConfigUtils.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.cqfn.save.core.plugin.PluginException
+import org.cqfn.save.plugin.warn.WarnPlugin
+import org.cqfn.save.plugins.fix.FixPlugin
+
+/**
+ * Evaluate all descendants of [this] config, reading individual plugin configurations from TOML file.
+ *
+ * @return [this] config with all descendants being evaluated (`pluginConfigs` are filled with data and merged with parents)
+ */
+fun TestConfig.withEvaluatedDescendantConfigsFromToml() = withEvaluatedDescendantConfigs {
```
kgevorkyan, line 18: May be rename these functions to something like `withDiscoveredDescendantConfigs` or `withObtainedDescendantConfigs`?

The `evaluate` seems to me somewhat inappropriate, what do you think?

petertrr: I think `discover` doesn't emphasize that configs are not simply discovered, but also filled with PuginConfigs and merged with parents. That's why I've chosen the word `evaluate`, to emphasize that this is sort of complicated configs traversal

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
petertrr: Good point about going twice, I'll try to reconsider it. The reason for this was that in save-cloud we now have two steps: at first loop over all `GeneralConfig`s to get test suite names, and then loop over all active plugins to get a full list of tests.

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TestConfigUtils.kt
kgevorkyan: Okay, agree. How about `processed ` then?
Don't want to be picky, but `evaluate` seems to me like `analyze` which isn't exactly what we do

I can't find some better option for now, so if you will still disagree, let's leave it as is

save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
petertrr: Removed the loop from this method and moved the call into the loop in `Save.kt`. I'll call `getAllTestConfigs().forEach` in save-cloud too

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TestConfigUtils.kt
petertrr: Changed the name after refactoring about the previous comment

kgevorkyan (approved):
lgtm

akuleshov7 (approved):

PR #140, opened 2021-06-21
Bugfix. Use save.properties from pwd as default value

kgevorkyan:
### What's done:
* Provide correct logic of usage of config options, in priority: CLI -> save.properties -> default values
* Make testFiles from SaveProperties nullable and engage in merge process (it was absent)


kgevorkyan:
Explanation could be found [here](https://github.com/cqfn/save/issues/125#issuecomment-865867600)

akuleshov7 (approved):

petertrr (approved):

PR #142, opened 2021-06-22
Support batchMode for a FixPlugin

### LINKED ISSUE #121: Support batchMode for a FixPlugin

Cheshiriks:
What's done:
* Added batchSize for a FixPlugin
* Added test
Closes #121


petertrr:
@Cheshiriks we still need general docs about batch mode

save-plugins/fix-plugin/src/commonTest/kotlin/org/cqfn/save/plugins/fix/FixPluginTest.kt
```diff
@@ -97,6 +97,52 @@ class FixPluginTest {
...
     }
 
+    @Test
+    @Suppress("TOO_LONG_FUNCTION")
+    fun `test for batchSize`() {
+        val config = fs.createFile(tmpDir / "save.toml")
+        val testFile1 = fs.createFile(tmpDir / "Test3Test.java")
+        val testFile2 = fs.createFile(tmpDir / "Test4Test.java")
+        fs.write(testFile1) {
+            write("Expected file".encodeToByteArray())
```
kgevorkyan, line 107: Did you mean `Original file`?
Why test files and expected files contain the same content before FixPlugin execution?

Cheshiriks: Fixed

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -48,38 +48,50 @@ class FixPlugin(
...
-                    return@map TestResult(
-                        listOf(expected, test),
-                        Fail("${ex::class.simpleName}: ${ex.message}", ex::class.simpleName ?: "Unknown exception"),
-                        DebugInfo(null, ex.message, null)
-                    )
-                }
-                val stdout = executionResult.stdout.joinToString("\n")
-                val stderr = executionResult.stderr.joinToString("\n")
 
+        return files.chunked(fixPluginConfig.batchSize ?: 1).map { chunk ->
```
petertrr, line 52:
```suggestion
        return files.chunked(fixPluginConfig.batchSize!!).map { chunk ->
```
Since batchSize already has a default value from `validateAndSetDefaults` method, there is no need to repeat it here. We can assume that the contract is, that config will have all required fields properly set by this time

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -48,38 +48,50 @@ class FixPlugin(
...
+            val pathMap = chunk.map { it.first() to it.last() }
+            val pathCopyMap = pathMap.map { (expected, test) -> expected to createTestFile(test) }
+            val testCopyNames = pathCopyMap.joinToString { (_, testCopy) -> testCopy.toString() }
```
petertrr, line 55: I'm concerned about separator for file names, both here and in warn-plugin. I think usually multiple CLI arguments are either separated by a whitespace or comma, not `", "`, like the default separator. I think we should make it configurable, probably in a separate PR.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -48,38 +48,50 @@ class FixPlugin(
...
+
+            val execCmd = "${(generalConfig!!.execCmd)} ${fixPluginConfig.execFlags} $testCopyNames"
+            val executionResult = try {
+                pb.exec(execCmd, null)
+            } catch (ex: ProcessExecutionException) {
+                return@map listOf(TestResult(
+                    pathMap.map { (expected, test) -> listOf(expected, test) }.flatten(),
```
petertrr, line 62:
```suggestion
                return@map chunk.map {
                    TestResult(it,
```
I think an additional `pathMap` collection is redundant, because `chunk` already is a list of groups of files,

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -48,38 +48,50 @@ class FixPlugin(
...
                 val expectedLines = FileSystem.SYSTEM.readLines(expected)
-                val status = diff(expectedLines, fixedLines).let { patch ->
-                    if (patch.deltas.isEmpty()) {
-                        Pass(null)
-                    } else {
-                        Fail(patch.formatToString(), patch.formatToShortString())
-                    }
-                }
+
+                val test = pathMap.filter { (_, test) -> test.name == testCopy.name }.map { (_, test) -> test }.first()
```
petertrr, line 75:
```suggestion
                val test = pathMap.first { (_, test) -> test.name == testCopy.name }.second
```

petertrr (changes requested):

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
Cheshiriks: Yes, I think the separator should be customizable, too

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
Cheshiriks: Fixed

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
petertrr: #158

petertrr (approved):

PR #148, opened 2021-06-23
Use config path relative to properties file location

### LINKED ISSUE #144: In Save.kt, `fullPathToConfig` should be built upon path to properties
I would expect that if I run SAVE from arbitrary location, providing `--property-file` should be enough, but now it looks for toml relative to pwd

petertrr:
### What's done:
* Changed logic to use config path relative to properties file location
* Generate file with own version
* Added welcome message with version

Closes #144

### DIFF
--- a/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
+++ b/save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
@@ -13,13 +13,11 @@ import org.cqfn.save.core.logging.logInfo
 import okio.FileNotFoundException
 import okio.FileSystem
 import okio.IOException
-import okio.Path.Companion.DIRECTORY_SEPARATOR
 import okio.Path.Companion.toPath
 
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.properties.Properties
 import kotlinx.serialization.serializer
-import okio.Path
 
 /**
  * @return this config in case we have valid configuration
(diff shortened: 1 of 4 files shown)


save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
```diff
@@ -24,13 +23,12 @@ import kotlinx.serialization.serializer
  * @return this config in case we have valid configuration
  */
 fun SaveProperties.validate(): SaveProperties {
-    val fullConfigPath = testRootPath + DIRECTORY_SEPARATOR + testConfigPath
+    this.testRootPath ?: logErrorAndExit(ExitCodes.INVALID_CONFIGURATION,
+        "`testRootPath` option is missing or null. " +
+                "Save is not able to start processing without an information about the tests that should be run.")
+    val fullConfigPath = propertiesFile!!.toPath().parent!! / testRootPath!! / testConfigPath!!
```
kgevorkyan, line 29: Hm, compiler couldn't identify, that `testRootPath` actually couldn't be null here?

kgevorkyan: Also, from description of `testRootPath`:
`"Path to directory with tests (relative path from place, where save.properties is stored or absolute path)"`

I think, now we don't support absolute path, right?
May be you also could change description in `config-options.json`, in aim to avoid misunderstandings

petertrr: > Hm, compiler couldn't identify, that testRootPath actually couldn't be null here?

I have actually tried it, but smart cast is unavailable here, because `testRootPath` is a mutable property

> I think, now we don't support absolute path, right?

Right, I think it could be supported relatively easy

kgevorkyan: Okay, let's keep it as is. I opened new issue about absolute path https://github.com/cqfn/save/issues/149

kgevorkyan (approved):
lgtm

PR #151, opened 2021-06-28
Minor changes in examples and PlainTextReporter

petertrr:
### What's done:
* Run save with default value for properties file in run.sh
* Introduced short reason for Fail
* Moved logic on test resources emptiness check into base Plugin class
* Collapse comment to one line in PlainTextReporter
* Improvements for kotlin-diktat example case
* Logging cleanup

Now PlainTextReporter ouputs like this:
```
--------------------------------
| Test name | result | comment |
| fix\smoke\Example1Expected.kt | Fail | ChangeDelta, position 0, lines:... |
--------------------------------
```


akuleshov7 (approved):

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -63,7 +58,7 @@ class FixPlugin(
                 } catch (ex: ProcessExecutionException) {
                     return@map TestResult(
                         listOf(expected, test),
-                        Fail(ex.message!!),
+                        Fail("${ex::class.simpleName}: ${ex.message}", ex::class.simpleName ?: "Unknown exception"),
```
kgevorkyan, line 61: What's the point to use just the class name of exception as `shortReason`? Also `ProcessExecutionException` is not anonymous object, so I think `simpleName` always exists

petertrr: Yes, this makes sense. Actually, I'm surprised, that pure Kotlin doesn't have Java-style toString for Throwables, with class and message. I've added it in https://github.com/cqfn/save/pull/154/commits/934e5ea9c41f6cec319ec51c117b41efab8c7175

PR #154, opened 2021-06-29
Bugfix: run individual tests and individual test suites

### LINKED ISSUE #152: Execution of individual tests is broken
`./save  warn/chapter1/save.toml` - is broken
`./save EnumValueSnakeCaseTest.kt` - is broken

petertrr:
### What's done:
* Changed logic

We have introduced a regression at some point, that both these modes have stopped working. This PR re-introduces this logic.

#### TODO:
* add tests - can be done under #155 

Closes #152

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
@@ -75,19 +75,21 @@ data class TestConfig(
             listOf(this) + this.childConfigs.flatMap { it.getAllTestConfigs() }
 
     /**
-     * recursively return all configs from the configuration Tree, belonging to [testFiles]
+     * recursively return all configs from the configuration Tree, belonging to [requestedConfigs]
      *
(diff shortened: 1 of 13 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/config/TestConfig.kt
```diff
@@ -67,13 +69,30 @@ data class TestConfig(
...
-        return listOf(this) + this.childConfigs.flatMap { it.getAllTestConfigs() }
-    }
+    fun getAllTestConfigsForFiles(requestedConfigs: List<String>): List<TestConfig> =
+            if (requestedConfigs.isEmpty()) {
+                getAllTestConfigs()
+            } else {
+                getAllTestConfigs().filter { testConfig ->
+                    requestedConfigs.any {
+                        // `testConfig`'s directory is a parent of `requestedConfig`'s directory
+                        testConfig.directory in it.toPath().parents() ||
```
akuleshov7, line 90: `it.toPath().parents()` move to variable? ot may be not

akuleshov7 (approved):

PR #163, opened 2021-06-30
Build optimizations

petertrr:
### What's done:
* Exclude detekt from `check` task
* Exclude linking release executable from ordinary CI workflow
* Enable gradle caches by default
* Disable redundant KMP targets compilation during build_and_test.yml
* Re-enable gradle-cache-action for build_and_test.yml


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DetektConfiguration.kt
```diff
@@ -21,6 +22,18 @@ fun Project.configureDetekt() {
         config = rootProject.files("detekt.yml")
         buildUponDefaultConfig = true
     }
+    // extremely awful hack to get to the point, when `DetektMultiplatform` has registered all the tasks
+    afterEvaluate {
+        afterEvaluate {
```
kgevorkyan, line 27: Could you please provide little comments about why we need such number of `afterEvaluate` calls?

kgevorkyan (approved):

Issue #160, opened 2021-06-30
warn-plugin: Support placeholders for line number


petertrr:
For example, `$LINE` should be resolved as current line and placeholder should support addition and subtraction:
```java
class example {
    // warn:{$LINE-1} [ClassNameIncorrect]
}
```
should be parsed as `Warning(line=1)`. The name of placeholder should be configurable (e.g. `%line`, `$l` etc.), and the whole expression should be configured using a capture group in regex

Issue #159, opened 2021-06-30
warn-plugin: Support default value for line number


petertrr:
Default value for line number should be equal to the next line:
```java
// warn [ClassNameIncorrect]
class example {}
```
should be parsed as `Warning(line=1, col=null ...)`
This functionality should be opt-in with config option

Issue #133, opened 2021-06-10
Support ignoreLines for all plugins: warn, fix, fix and warn


petertrr:
Fix plugin should remove comments when creating a copy; this way users can write expected files without these comments. This is the case when user wants to coimbine warn and fix plugins on same resourceds.

PR #165, opened 2021-07-01
Implement FixAndWarn plugin

kgevorkyan:
### What's done:
* Implement new plugin: FixAndWarn plugin, which combines two actions during one execution: fix and warn
* Unify common logic in creation of copy of tests files in plugin into one method 
* Add REAMDE
* Add test


save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -0,0 +1,181 @@
...
+        // Remove (in place) warnings from test files before fix plugin execution
+        val filesAndTheirWarningsMap = removeWarningsFromExpectedFiles(expectedFiles)
+
+        val fixTestResults = fixPlugin.handleFiles(files)
+
+        // Fill back original data with warnings
+        filesAndTheirWarningsMap.forEach { (filePath, warningsList) ->
+            val fileData = fs.readLines(filePath) as MutableList
+            warningsList.forEach { (line, warningMsg) ->
+                fileData.add(line, warningMsg)
```
petertrr, line 90: So ,you simply append comments with warnings to the end of the file? What if comments have dynamically calculated line number, like we will support in #159 and #160 ?
We have a related issue #133. and logic can be implemented for both Fix and FixAndWarn plugins there. Maybe, we should fix file *with* comments and then remove them when *comparing* with expected? This way we can directly check fixed file with warn-plugin, and if they use placeholders for line number, they don't need to keep in mind line changes.

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -0,0 +1,181 @@
...
+     * Filter test resources
+     *
+     * @param suffix regex pattern of test resource
+     * @param match whether to keep elements which matches current pattern or to keep elements, which is not
+     * @return filtered list of files
+     */
+    private fun Sequence<List<Path>>.filterTestResources(suffix: Regex, match: Boolean): List<Path> {
+        val filteredFiles: MutableList<Path> = mutableListOf()
+        this.forEach { resources ->
+            filteredFiles.add(resources.single { path ->
```
petertrr, line 123: Same question about `map` here

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -0,0 +1,181 @@
...
+     */
+    private fun removeWarningsFromExpectedFiles(files: List<Path>): MutableMap<Path, WarningsList> {
+        val filesAndTheirWarningsMap: MutableMap<Path, WarningsList> = mutableMapOf()
+        files.forEach { file ->
+            val fileData = fs.readLines(file)
+            filesAndTheirWarningsMap[file] = mutableListOf()
+
+            val fileDataWithoutWarnings = fileData.filterIndexed { index, line ->
+                val isLineWithWarning = (warnPluginConfig.warningsInputPattern!!.find(line)?.groups != null)
+                if (isLineWithWarning) {
```
petertrr, line 149: You can chain `filterIndexed` and `onEach` to make it more clearly looking

petertrr (changes requested):

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
kgevorkyan: Actually, not really. `warningsList` contain the pairs of line number of warning and warning message from the original version of expected file.
So, since in `fileData` there is a content of expected file without any warnings, and also we know (by warningsList) on which lines there were warnings, we just add them into appropriate place. Here I used the overload of `add` function, which `Inserts an element into the list at the specified index`.

About `What if comments have dynamically calculated line number` - I think it is not connected with current algorithm, since here I save the line not of problematic code on which we should raise warn, but the line of warning itself.

Don't really understand how we can check for expected warnings if user won't provide them to us `users can write expected files without these comments`.

petertrr: Oh, didn't see, that it's `add` with index, sorry.

> Don't really understand how we can check for expected warnings if user won't provide them to us users can write expected files without these comments.

I mean, that only `*Test` files should contain comments with warnings, because we can implement comparison with expected files with these comments removed. This will allow users not to write same comments twice. And warn-plugin needs to operate on mutated `*Test` file anyway.

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -0,0 +1,181 @@
...
+
+        // Test files for warn plugin could be obtained by filtering of `fix` resources: excluding `expected` files
+        val testFilePattern = warnPluginConfig.resourceNamePattern
+        val warnTestFiles = files.filterTestResources(testFilePattern, match = true)
+
+        // Warn plugin should process the files, which were fixed by fix plugin
+        // We can get the paths of fixed files from warn test files (since they are the same for fix plugin),
+        // and then just add relative path from FixPlugin tmp dir
+        val testFilesAfterFix: MutableList<List<Path>> = mutableListOf()
+        warnTestFiles.forEach { path ->
```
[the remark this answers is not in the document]
kgevorkyan: Changed to map

save-plugins/fix-and-warn-plugin/README.md
```diff
@@ -0,0 +1,41 @@
...
+Plugin fix test file, warns if something couldn't be auto-corrected after fix
+and compares output with expected output during one execution.
+
+There are no additional settings, which should be provided for this plugin,
+it just should contain subsections `[fix]` and `[warn]`, which should be configured,
+like it described in corresponding README files.
+
+Please note, that it is only important for test resources to have the same postfixes in 
+`[fix]` and `[warn]` sections.
+By the default test file it is `Test`,
```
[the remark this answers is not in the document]
kgevorkyan: Fixed

save-plugins/fix-and-warn-plugin/README.md
```diff
@@ -0,0 +1,41 @@
+## Save fix-and-warn plugin
+Plugin that runs provided executable on the initial file with a test source code
+and combines two actions during one execution: 
+[fix](save-plugins/fix-plugin/README.md) and [warn](save-plugins/warn-plugin/README.md).\
+Plugin fix test file, warns if something couldn't be auto-corrected after fix
```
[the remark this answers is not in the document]
kgevorkyan: Thanks, fixed

save-plugins/fix-and-warn-plugin/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+        all {
+            languageSettings.useExperimentalAnnotation("kotlin.RequiresOptIn")
+            languageSettings.useExperimentalAnnotation("okio.ExperimentalFileSystem")
+        }
+        val commonMain by getting {
+            dependencies {
+                implementation(project(":save-common"))
+                implementation(project(":save-plugins:fix-plugin"))
+                implementation(project(":save-plugins:warn-plugin"))
+                implementation("io.github.petertrr:kotlin-multiplatform-diff:0.2.0")
```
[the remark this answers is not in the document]
kgevorkyan: Done

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
kgevorkyan: Discussed offline. For now warn plugin will check warnings in `Expected` files

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
kgevorkyan: Changed to map

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
kgevorkyan: Seems it is not very suitable for current situation

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -0,0 +1,161 @@
...
+    override fun handleFiles(files: Sequence<List<Path>>): Sequence<TestResult> {
+        testConfig.validateAndSetDefaults()
+
+        val testFilePattern = warnPluginConfig.resourceNamePattern
+        val expectedFiles = files.filterTestResources(testFilePattern, match = false)
+
+        // Remove (in place) warnings from test files before fix plugin execution
+        val filesAndTheirWarningsMap = removeWarningsFromExpectedFiles(expectedFiles)
+
+        val fixTestResults = fixPlugin.handleFiles(files)
```
petertrr, line 68: This is connected with #133: instead of 
- reading the file and overwriting it to store warnings in memory
- reading it once more to compare with fixed
- restoring original contents for warn-plugin
we could support ignoring comments inside fix-plugin: it would read expected files for comparison and filter out warnings.

petertrr (approved):
Let's use it as an initial implementation, we already have a few opened issues to improve and enhance it

akuleshov7:
Could you please add example with plugin to our EXAMPLES folder

akuleshov7 (approved):

kgevorkyan:
> Could you please add example with plugin to our EXAMPLES folder

We have separate issue for this https://github.com/cqfn/save/issues/170, I think it could be done in future PR

PR #166, opened 2021-07-02
Implement json reporter

### LINKED ISSUE #146: Reporters: Implement JSON reporter
Could be as simple as deserialization of TestResults into strings and writing them into a Sink

petertrr:
### What's done:
* Fixed reporter calls in Save.kt
* Remove `onPluginExcutionSkip` from `Reporter` and added debug logging in `buildActivePlugins`
* Added Report.kt
* Added JsonReporter.kt
* Added serialization for Crashes
* Added `MockPlugin`
* Added tests

`Report` can be used in save-cloud to deserialize reports from SAVE output.

`MockPlugin` can be enhanced in the future, e.g. it can accepts `TestResult`s and emit them, when `execute` is called.

Closes #146


save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -103,13 +107,20 @@ class Save(
     }
 
     private fun getReporter(saveProperties: SaveProperties): Reporter {
+        val outFileName = when (saveProperties.reportType!!) {
+            ReportType.PLAIN -> "save.out"
```
kgevorkyan, line 111: May be we can create separate variable instead of hardcoded `save.out` in `ReportType`? And here operate just with expansions?

save-reporters/src/commonMain/kotlin/org/cqfn/save/reporter/json/JsonReporter.kt
```diff
@@ -0,0 +1,82 @@
...
+    override fun afterAll() {
+        out.write("]\n".encodeToByteArray())
+    }
+
+    override fun onSuiteStart(suiteName: String) {
+        if (!isFirstSuite) {
+            out.write(",\n".encodeToByteArray())
+        } else {
+            isFirstSuite = false
+        }
```
kgevorkyan, line 44: Can we move this part into separate function?
It's duplicates in each method

save-common/src/commonMain/kotlin/org/cqfn/save/core/result/TestStatus.kt
```diff
@@ -4,30 +4,37 @@
...
 
 /**
  * Represents the case when test execution crashed because of an unhandled internal error in SAVE framework.
  *
- * @property throwable an exception that caused crash of SAVE
+ * @property description description of an exception that caused crash of SAVE
  */
-data class Crash(val throwable: Throwable) : TestStatus()
+@Serializable
+data class Crash(val description: String) : TestStatus()
```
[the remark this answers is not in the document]
petertrr: When `Crash` is created, I use output of `Throwab;e.describe()`. But it makes more sense to keep them as separate fiedls, I'll fix it.

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
petertrr: Extracted it into a variable for now, we should make it configurable via cli/properties in the future

save-reporters/src/commonMain/kotlin/org/cqfn/save/reporter/json/JsonReporter.kt
petertrr: Yeah, tried to refactor it. Looks a little weird, because boolean flag can't be changed from inside the function. When we use atomic, it will look better.

save-core/src/commonMain/kotlin/org/cqfn/save/core/Save.kt
kgevorkyan: Ok, let's open corresponding issue on this subject

petertrr: #169

kgevorkyan (approved):

PR #167, opened 2021-07-05
Separator for batchMode should be configurable

### LINKED ISSUE #158: Separator for batchMode should be configurable
I'm concerned about separator for file names, both here and in warn-plugin. I think usually multiple CLI arguments are either separated by a whitespace or comma, not `", "`, like the default separator. I think we should make it configurable, probably in a separate PR.

_Originally posted by @petertrr in https://github.com/cqfn/save/pull/142#discussion_r660598904_

Cheshiriks:
What's done:
* Added separator config for a FixPlugin and WarnPlugin
Closes #158

### DIFF
--- a/save-plugins/fix-plugin/README.md
+++ b/save-plugins/fix-plugin/README.md
@@ -47,13 +47,13 @@ execFlags="-F"
 testFilePattern="*Test.kt"
 expectedFilePattern="*Expected.kt"
 batchSize = 1 # (default value)
-separator = ", " # (default value)
+batchSeparator = ", " # (default value)
 resourceNameTestSuffix = "Test" # (default value)
 resourceNameExpectedSuffix = "Expected" #(default value)
 ```
 
 When executed from project root (where `save.propertes` is located), SAVE will cd to `rootDir` and discover all pairs of files
 matching `testFilePattern` and `expectedFilePattern` with same prefix. It will then execute `$execCmd $testFile`. `batchSize` it controls how many files execCmd will process at a time. (since we specified
-`batchSize = 1`, it will provide inputs one by one) and compare its stdout (as per `output` option) with respecting `$expectedFile`. `separator` is a setting for batchSize.
+`batchSize = 1`, it will provide inputs one by one) and compare its stdout (as per `output` option) with respecting `$expectedFile`.  `batchSeparator` is separator for filenames in `execCmd` if `batchSize > 1`.
 Results will be written in plain text as well as JSON.
 `resourceNameTestSuffix` must include suffix name of the test file. `resourceNameExpectedSuffix` must include suffix name of the expected file.
--- a/save-plugins/fix-plugin/README.md
+++ b/save-plugins/fix-plugin/README.md
@@ -46,13 +46,14 @@ suiteName = "DocsCheck"
 execFlags="-F"
 testFilePattern="*Test.kt"
 expectedFilePattern="*Expected.kt"
-batchSize = 1
+batchSize = 1 # (default value)
+separator = ", " # (default value)
 resourceNameTestSuffix = "Test" # (default value)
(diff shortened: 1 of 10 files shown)


save-plugins/fix-plugin/README.md
```diff
@@ -46,13 +46,14 @@ suiteName = "DocsCheck"
 execFlags="-F"
 testFilePattern="*Test.kt"
 expectedFilePattern="*Expected.kt"
-batchSize = 1
+batchSize = 1 # (default value)
+separator = ", " # (default value)
```
petertrr, line 50:
```suggestion
batchSeparator = ", " # (default value, not used if `batchSize = 1`)
```
As these two settings are connected, they should have similar names

save-plugins/fix-plugin/README.md
```diff
@@ -46,13 +46,14 @@ suiteName = "DocsCheck"
...
 resourceNameTestSuffix = "Test" # (default value)
 resourceNameExpectedSuffix = "Expected" #(default value)
 ```
 
 When executed from project root (where `save.propertes` is located), SAVE will cd to `rootDir` and discover all pairs of files
 matching `testFilePattern` and `expectedFilePattern` with same prefix. It will then execute `$execCmd $testFile`. `batchSize` it controls how many files execCmd will process at a time. (since we specified
-`batchSize = 1`, it will provide inputs one by one) and compare its stdout (as per `output` option) with respecting `$expectedFile`.
+`batchSize = 1`, it will provide inputs one by one) and compare its stdout (as per `output` option) with respecting `$expectedFile`. `separator` is a setting for batchSize.
```
petertrr, line 57: Please add more description: which requirements of `execCmd` can be satisfied by changing `separator`?

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -54,7 +54,7 @@ class FixPlugin(
         return files.chunked(fixPluginConfig.batchSize!!).map { chunk ->
             val pathMap = chunk.map { it.first() to it.last() }
             val pathCopyMap = pathMap.map { (expected, test) -> expected to createTestFile(test) }
-            val testCopyNames = pathCopyMap.joinToString { (_, testCopy) -> testCopy.toString() }
+            val testCopyNames = pathCopyMap.joinToString(separator = fixPluginConfig.separator!!) { (_, testCopy) -> testCopy.toString() }
```
petertrr, line 57: Is this function also called if `batchSize == 1`? If so, we should correctly handle nullable separator in that case

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -101,7 +101,7 @@ class WarnPlugin(
             )
         }
 
-        val fileNames = paths.joinToString {
+        val fileNames = paths.joinToString(separator = warnPluginConfig.separator!!) {
```
petertrr, line 104: Same here about batchSize == 1

petertrr (changes requested):

save-plugins/fix-plugin/README.md
Cheshiriks: Fixed

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
petertrr: Despite being nullable, it always has default value, so no need to check additionally

petertrr (approved):

PR #171, opened 2021-07-07
Running individual tests doesn't work as expected

### LINKED ISSUE #137: Running individual tests doesn't work as expected
* [x] ~`saveProperties.testFiles` should be passed to plugins on creation in `Save.kt`~ - fixed in #154
* [ ] For individual tests, only file name match is currently checked; it would be better to check the whole path
* [x] ~Running tests in individual test suite like `save -prop save.properties tests/suite1/save.toml` doesn't work~ - fixed in #154

Cheshiriks:
What's done:
* Check for the full path to the file
Closes #137

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -63,7 +63,9 @@ abstract class Plugin(
         return if (testFiles.isNotEmpty()) {
             rawTestFiles.filter { paths ->
                 // test can be specified by the name of one of it's files
-                paths.any { it.name in testFiles }
+                paths.any { path ->
+                    testFiles.any { it in path.toString() }
+                }
             }
         } else {
             rawTestFiles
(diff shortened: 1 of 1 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -63,7 +63,9 @@ abstract class Plugin(
         return if (testFiles.isNotEmpty()) {
             rawTestFiles.filter { paths ->
```
petertrr, line 64:
```suggestion
            rawTestFiles.filter { resourcesGroup ->
```

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -63,7 +63,9 @@ abstract class Plugin(
...
                 // test can be specified by the name of one of it's files
-                paths.any { it.name in testFiles }
+                paths.any { path ->
+                    testFiles.any { it in path.toString() }
```
petertrr, line 67: Do we have any tests or at least docs on this mode? Please update them, or add into the main readme, if they are missing.

petertrr (changes requested):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
Cheshiriks: Yes. In main readme in "Executing specific tests" indicates that you need to pass the path to the test file.

petertrr (approved):

Issue #170, opened 2021-07-07
Expand examples for FixAndWarnPlugin


kgevorkyan:
Examples should expanded with test suites, on which FixAndWarn plugin will be executed

PR #174, opened 2021-07-12
Update ktoml version and provide corresponding changes in plugin configs

kgevorkyan:
### What's done:
* Update ktoml version
* Refactor TomlUtils according library API
* Cosmetic changes

### DIFF
--- a/save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
+++ b/save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
@@ -6,18 +6,41 @@ package org.cqfn.save.core.utils
 
 import org.cqfn.save.core.config.TestConfigSections
 import org.cqfn.save.core.logging.logError
+import org.cqfn.save.core.plugin.GeneralConfig
 import org.cqfn.save.core.plugin.PluginConfig
 import org.cqfn.save.core.plugin.PluginException
+import org.cqfn.save.plugin.warn.WarnPluginConfig
 import org.cqfn.save.plugins.fix.FixPluginConfig
+import org.cqfn.save.plugins.fixandwarn.FixAndWarnPluginConfig
 
 import com.akuleshov7.ktoml.KtomlConf
 import com.akuleshov7.ktoml.deserializeTomlFile
 import com.akuleshov7.ktoml.exceptions.KtomlException
 import com.akuleshov7.ktoml.parsers.TomlParser
+import com.akuleshov7.ktoml.parsers.node.TomlTable
 import okio.Path
 
 import kotlinx.serialization.ExperimentalSerializationApi
-import kotlinx.serialization.serializer
+
+private fun Path.testConfigFactory(table: TomlTable) =
+        when (table.fullTableName.uppercase()) {
+            TestConfigSections.FIX.name -> this.createPluginConfig<FixPluginConfig>(
+                table.fullTableName
+            )
+            TestConfigSections.`FIX AND WARN`.name -> this.createPluginConfig<FixAndWarnPluginConfig>(
+                table.fullTableName
+            )
+            TestConfigSections.WARN.name -> this.createPluginConfig<WarnPluginConfig>(
+                table.fullTableName
+            )
+            TestConfigSections.GENERAL.name -> this.createPluginConfig<GeneralConfig>(
+                table.fullTableName
+            )
+            else -> throw PluginException(
+                "Received unknown plugin section name in the input: [${table.fullTableName}]." +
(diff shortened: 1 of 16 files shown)


save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
```diff
@@ -56,45 +50,8 @@ private inline fun <reified T : PluginConfig> Path.createPluginConfig(
...
-        configList.add(sectionPluginConfig)
-    }
-
-    return configList
-}
+fun createPluginConfigListFromToml(testConfigPath: Path): List<PluginConfig> =
+    TomlParser(KtomlConf())
+        .readAndParseFile(testConfigPath.toString())
+        .getRealTomlTables()
+        .map { testConfigPath.createPluginConfig<FixPluginConfig>(it.name.uppercase()) }
```
kgevorkyan: Why just `FixPluginConfig`?
What about other plugin configs?

akuleshov7: oh, fuck

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
```diff
@@ -56,45 +51,23 @@ private inline fun <reified T : PluginConfig> Path.createPluginConfig(
...
-            )
-            TestConfigSections.WARN.name -> testConfigPath.createPluginConfig<WarnPluginConfig>(
-                fakeFileNode,
-                sectionName
-            )
-            TestConfigSections.GENERAL.name -> testConfigPath.createPluginConfig<GeneralConfig>(
-                fakeFileNode,
-                sectionName
-            )
-            else -> throw PluginException(
```
kgevorkyan, line 90: I think we can hold this branch in new version of `TomlUtils` too

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
```diff
@@ -56,45 +51,23 @@ private inline fun <reified T : PluginConfig> Path.createPluginConfig(
...
+fun createPluginConfigListFromToml(testConfigPath: Path): List<PluginConfig> =
+    TomlParser(KtomlConf())
+        .readAndParseFile(testConfigPath.toString())
+        .getRealTomlTables()
+        .map {
+            when (TestConfigSections.valueOf(it.fullTableName.uppercase())) {
+                TestConfigSections.FIX -> testConfigPath.createPluginConfig<FixPluginConfig>(
+                    it.fullTableName
+                )
+                TestConfigSections.`FIX AND WARN` -> testConfigPath.createPluginConfig<FixAndWarnPluginConfig>(
```
kgevorkyan, line 63: Could you please update name of this section in corresponding README of `FIX AND WARN` plugin

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
akuleshov7: The exception will be thrown on the parsing of enum in case of an invalid  I do not think that we really need ‘else’ branch. That is a syntax sugar by kotlin

akuleshov7: Why we even just throw an exception? Where we handle this result? This will simply stop the test…

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
akuleshov7: How did it work before? In readme you had [fix-and-warn], in Enum you had FIX_AND_WARN

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
akuleshov7: First I thought that it is not needed, but yes, logging becomes much better with this

akuleshov7 (approved):

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
kgevorkyan: For this moment we have no examples for it https://github.com/cqfn/save/issues/170
However there is exist unit test in `commonTest` directory for this plugin

PR #175, opened 2021-07-12
warn-plugin: Support default value for line number

### LINKED ISSUE #159: warn-plugin: Support default value for line number

Cheshiriks:
What's done:
* Added defaultLineMode in warn plugin
* Added test
Closes #159

### DIFF
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
@@ -14,6 +14,7 @@ import org.cqfn.save.core.result.TestStatus
 import org.cqfn.save.core.utils.ProcessExecutionException
 import org.cqfn.save.plugin.warn.utils.Warning
 import org.cqfn.save.plugin.warn.utils.extractWarning
+import org.cqfn.save.plugin.warn.utils.getLineNumber
 
 import okio.FileSystem
 import okio.Path
@@ -69,7 +70,7 @@ class WarnPlugin(
         warningRegex: Regex,
         linesFile: List<String>,
         lineNum: Int): Int {
-        var x = 0
+        var x = 1
         while (lineNum - 1 + x < linesFile.size && warningRegex.find(linesFile[lineNum - 1 + x]) != null) {
             x++
         }
@@ -86,25 +87,39 @@ class WarnPlugin(
         generalConfig: GeneralConfig?): List<TestResult> {
         val expectedWarnings: WarningMap = mutableMapOf()
         paths.forEach { path ->
-            var i = 1
             val linesFile = fs.readLines(path)
             expectedWarnings.putAll(
-                fs.readLines(path)
-                    .map {
-                        it to plusLine(warnPluginConfig.warningsInputPattern!!, linesFile, i++)
-                    }
-                    .mapNotNull {
-                        with(warnPluginConfig) {
-                            it.first.extractWarning(
-                                defaultLineMode!!,
-                                warningsInputPattern!!,
-                                path.name,
-                                if (defaultLineMode) it.second else lineCaptureGroup,
-                                columnCaptureGroup,
-                                messageCaptureGroup!!,
-                            )
+                if (warnPluginConfig.defaultLineMode!!) {
+                    linesFile
+                        .mapIndexed { index, line ->
+                            line to plusLine(warnPluginConfig.warningsInputPattern!!, linesFile, index)
                         }
-                    }
+                        .mapNotNull {
+                            with(warnPluginConfig) {
+                                it.first.extractWarning(
+                                    warningsInputPattern!!,
+                                    path.name,
+                                    it.second,
+                                    columnCaptureGroup,
+                                    messageCaptureGroup!!,
+                                )
+                            }
+                        }
+                } else {
+                    linesFile
+                        .mapNotNull {
+                            with(warnPluginConfig) {
+                                val line = it.getLineNumber(warningsInputPattern!!, lineCaptureGroup)
+                                it.extractWarning(
+                                    warningsInputPattern,
+                                    path.name,
+                                    line,
+                                    columnCaptureGroup,
+                                    messageCaptureGroup!!,
+                                )
+                            }
+                        }
+                }
                     .groupBy {
                         if (it.line != null && it.column != null) {
                             it.line to it.column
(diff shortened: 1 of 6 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -75,14 +86,20 @@ class WarnPlugin(
         generalConfig: GeneralConfig?): List<TestResult> {
         val expectedWarnings: WarningMap = mutableMapOf()
         paths.forEach { path ->
+            var i = 1
+            val linesFile = fs.readLines(path)
             expectedWarnings.putAll(
                 fs.readLines(path)
```
petertrr, line 92:
```suggestion
                linesFile
```
Why read file twice?

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -75,14 +86,20 @@ class WarnPlugin(
         generalConfig: GeneralConfig?): List<TestResult> {
         val expectedWarnings: WarningMap = mutableMapOf()
         paths.forEach { path ->
+            var i = 1
```
petertrr, line 89: You should use `mapIndexed` instead of such counters

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -65,6 +65,17 @@ class WarnPlugin(
         }
     }
 
+    private fun plusLine(
```
petertrr, line 68: So, this thing returns the number of the next line without warnings after `lineNum`?

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -75,14 +86,20 @@ class WarnPlugin(
         generalConfig: GeneralConfig?): List<TestResult> {
         val expectedWarnings: WarningMap = mutableMapOf()
         paths.forEach { path ->
+            var i = 1
+            val linesFile = fs.readLines(path)
             expectedWarnings.putAll(
                 fs.readLines(path)
+                    .map {
+                        it to plusLine(warnPluginConfig.warningsInputPattern!!, linesFile, i++)
```
petertrr, line 94: This should be calculated only if the new mode is active

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/utils/Warning.kt
```diff
@@ -102,5 +94,44 @@ internal fun String.extractWarning(warningRegex: Regex,
...
+ * @param messageGroupIdx index of capture group for waring text
+ * @return a [Warning] or null if [this] string doesn't match [warningRegex]
+ * @throws ResourceFormatException when parsing a file
+ */
+@Suppress(
+    "TOO_MANY_PARAMETERS",
+    "LongParameterList",
+    "TooGenericExceptionCaught",
+    "SwallowedException")
+internal fun String.extractWarning(defaultLineMode: Boolean,
```
petertrr, line 115: I'd rather add a new method `getLineNumber(defaultLineMode, lineGroupIdx, defaultLineNumber)` and call it from `mapNotNull` in `WarnPlugin`. And later this method can be enhanced with placeholders logic

petertrr (changes requested):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
Cheshiriks: Yes. This method checks if there is another warning after the warning or if there is a line with the code and returns its number.

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -65,6 +66,17 @@ class WarnPlugin(
         }
     }
 
+    private fun plusLine(
+        warningRegex: Regex,
+        linesFile: List<String>,
+        lineNum: Int): Int {
+        var x = 1
+        while (lineNum - 1 + x < linesFile.size && warningRegex.find(linesFile[lineNum - 1 + x]) != null) {
```
petertrr, line 74: What if `x` becomes larger than the number of lines? It means, that the user has written warnings in the end of the file, and there are no lines left to calculate line numbers. I believe in this case we should log a warning, while now you are silently setting the last line number

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -75,19 +87,39 @@ class WarnPlugin(
...
+                                    path.name,
+                                    it.second,
+                                    columnCaptureGroup,
+                                    messageCaptureGroup!!,
+                                )
+                            }
+                        }
+                } else {
+                    linesFile
+                        .mapNotNull {
```
petertrr, line 110: These two `mapNotNull` are pure copy-paste, you should reuse them in both modes

petertrr (changes requested):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -75,19 +93,39 @@ class WarnPlugin(
...
-                    .mapNotNull {
-                        with(warnPluginConfig) {
-                            it.extractWarning(
-                                warningsInputPattern!!,
-                                path.name,
-                                lineCaptureGroup,
-                                columnCaptureGroup,
-                                messageCaptureGroup!!,
-                            )
+                if (warnPluginConfig.defaultLineMode!!) {
```
petertrr, line 98: Maybe, we can make it like this:
```kotlin
linesFile.mapIndexed { index, line ->
    if (warnPlugin.defaultLineMode!!) {
        // extraact warning with `plusLine`
    } else {
        // extractWarning as usual
    }
}
.filterNotNull()
```
Because having a large if expression, followed by a bunch of other calls with lambdas makes it really overwhelming. And, AFAIU, we don't really need to iterate twice when using `plusLine`

Cheshiriks: Yes. Good idea:)
Fixed

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -95,37 +95,24 @@ class WarnPlugin(
...
+                                warningsInputPattern!!,
+                                path.name,
+                                newLine,
+                                columnCaptureGroup,
+                                messageCaptureGroup!!,
+                            )
                         }
-                }
+                    }
+                    .mapNotNull { it }
```
petertrr, line 115: `filterNotNull()`

petertrr (approved):

PR #181, opened 2021-07-16
Merge testRootPath and propertiesFile options into one

### LINKED ISSUE #141: Can testConfigPath and testRootPath from SaveProperties be merged?
Can testConfigPath and testRootPath from SaveProperties be merged? Now the only place where we use them is to get full path to config, which we can store in properties right away.

Originally posted in https://github.com/cqfn/save/issues/108

### LINKED ISSUE #179: Merge testRootPath and propertiesFile options into a single testRootDir
Related issues: #141, #149

petertrr:
### What's done:
* Update config-options.json, regenerated docs and SaveProperties
* Updated logic
* Bump tools versions in examples/kotlin-diktat
* Pass resources paths to reporter relative to testRootDir

Closes #179
Closes #141

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Save can be used not only with static analyzers, but can be used as a test frame
 
 ## How to start
 1. Prepare and configure your test base in the proper format. See [test_detection](#test_detection) and [plugins](#plugins)
-2. Run the following: `save --testConfig "/my/path/to/test/save.toml"`
+2. Run the following: `save --testRootPath "/my/path/to/tests"`. Directory `tests` should contain `save.toml` configuration file.
 
 ## How to configure 
 SAVE has a command line interface that runs the framework and your executable. What you need is simply to configure the output of your static analyzer so SAVE will be able to
@@ -57,16 +57,14 @@ To check that the warning is correct for SAVE - your static analyzer must print 
 
 // FixMe: specify options here
 For [save-cloud](https://github.com/cqfn/save-cloud) it is extremely important to detect properties file for SAVE,
-so in case you would like to use both cli application and cloud application we suggest putting the file with name `save.properties` in the root folder of your repository.
-
-Configuration file can be also provided to SAVE via command line: `save -prop $PATH_TO_PROPS/save.properties`
+so in case you would like to use both cli application and cloud application we suggest putting the file with name `save.properties`
+in the same folder as root test config `save.toml`.
 
 Example of `save.properties` file (for full description and possible values, refer to the [corresponding section](#config_options)):
 ```properties
 resultOutput=stdout
 parallelMode=true
 language=c++
-testRootPath=~/my_dir/dir_to_tests_root
 reportDir=~/report
 ```
 
--- a/README.md
+++ b/README.md
@@ -56,9 +56,7 @@ check if the proper error was raised on the proper line of test code.
 To check that the warning is correct for SAVE - your static analyzer must print the result to stderr/stdout or to some log file.
 
 // FixMe: specify options here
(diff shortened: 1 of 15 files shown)


save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
```diff
@@ -54,7 +54,7 @@ fun createConfigFromArgs(args: Array<String>): SaveProperties {
     val configFromCli = SaveProperties(args)
     logDebug("Properties after parsed command line args:\n${configFromCli.getFields()}")
     // reading configuration from the properties file
-    val configFromPropertiesFile = readPropertiesFile(configFromCli.propertiesFile ?: defaultConfig().propertiesFile)
+    val configFromPropertiesFile = readPropertiesFile("${(configFromCli.testRootPath ?: defaultConfig().testRootPath)}/save.properties")
```
kgevorkyan, line 57: It worked? linux and windows have different dir separators

```suggestion
    val configFromPropertiesFile = readPropertiesFile("${(configFromCli.testRootPath ?: defaultConfig().testRootPath)}${DIRECTORY_SEPARATOR}save.properties")
```

README.md
```diff
@@ -57,16 +57,14 @@ To check that the warning is correct for SAVE - your static analyzer must print
 
 // FixMe: specify options here
 For [save-cloud](https://github.com/cqfn/save-cloud) it is extremely important to detect properties file for SAVE,
-so in case you would like to use both cli application and cloud application we suggest putting the file with name `save.properties` in the root folder of your repository.
-
-Configuration file can be also provided to SAVE via command line: `save -prop $PATH_TO_PROPS/save.properties`
+so in case you would like to use both cli application and cloud application we suggest putting the file with name `save.properties`
```
kgevorkyan, line 60: We `suggest` or `demand`?

petertrr: I mean, we try to ask politely :) But actually, `save.properties` isn't that needed for cloud anymore, because we only needed this to discover root config. Now we can just pass path to test root dir and it will be enough for cloud. I'll change this sentence in readme

examples/kotlin-diktat/run.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 KTLINT_VERSION=0.39.0
-DIKTAT_VERSION=1.0.0-rc.1
-SAVE_VERSION=0.1.0-rc.1
+DIKTAT_VERSION=1.0.0-rc.2
```
kgevorkyan, line 4: Thoughts out loud:
I don't really like, that now we have such amount of places, where same versions are stored (`build.gradle.kts` from `buildSrc`, `Versions.kt` and 2 places in `examples`), and should keep in mind it after every release.

bash can [assign variable from another file](https://stackoverflow.com/questions/20406134/read-a-bash-variable-assignment-from-other-file), however I don't really like it.
May be, we can create something better in the future

save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
petertrr: Thanks, fixed. I only tested it in mingw, and actually it did work, even though separators in json were `/` instead of `\\`.

README.md
kgevorkyan: Ok, just thought, that we need some place in README, where we say to user, that for proper work, if `save.properties` is provided, it should be at the same dir of root toml config

petertrr: Yes, I've added a line about it

kgevorkyan (approved):

PR #185, opened 2021-07-20
Examples for FixAndWarn plugin

kgevorkyan:
### What's done:
* Provide examples for FixAndWarn plugin
* Little refactoring of logic in FixAndWarn plugin: fields, responsible for [fix] and [warn] subsections should be updated after validation
* Rename fields of FixAndWarnConfig according proper naming in toml file


save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
```diff
@@ -76,5 +76,6 @@ private inline fun <reified T : PluginConfig> Path.createPluginConfig(
 fun createPluginConfigListFromToml(testConfigPath: Path): List<PluginConfig> =
         TomlParser(KtomlConf())
             .readAndParseFile(testConfigPath.toString())
-            .getRealTomlTables()
+            // Current function returns all tables (including subtables), however we need to extract only top level sections
+            .getRealTomlTables().filter { it.level == 0 }
```
petertrr, line 80: Should this check be moved inside `getRealTomlTables`? Or into a separate method `getTopLevelTomlTables`?

save-plugins/fix-and-warn-plugin/src/commonTest/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPluginTest.kt
```diff
@@ -101,7 +101,8 @@ class FixAndWarnPluginTest {
             ), results.first())
         val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY / FixPlugin::class.simpleName!!)
         assertTrue("Files should be identical") {
-            diff(fs.readLines(tmpDir / "Test1Test.java"), fs.readLines(expectedFile))
+            // Additionally ignore warnings in expected file
+            diff(fs.readLines(tmpDir / "Test1Test.java"), fs.readLines(expectedFile).filter { !it.contains("warn") })
```
petertrr, line 105:
```suggestion
            diff(fs.readLines(tmpDir / "Test1Test.java"), fs.readLines(expectedFile).filterNot { it.contains("warn") })
```

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -85,11 +85,12 @@ class FixAndWarnPlugin(
         // TODO: then warn plugin should look at the fix plugin output for actual warnings, and not execute command one more time.
         // TODO: However it's required changes in warn plugin logic (it's should be able to compare expected and actual warnings from different places),
         // TODO: this probably could be obtained after https://github.com/cqfn/save/issues/164,
-        val warnTestResults = warnPlugin.handleFiles(expectedFiles.map { listOf(it) })
-        return fixTestResults + warnTestResults
+        val warnTestResults = warnPlugin.handleFiles(expectedFiles.map { listOf(it) }).toList()
+        return fixTestResults.asSequence() + warnTestResults.asSequence()
```
petertrr, line 89: Why do you need `toList` and then `asSequence`? Sequence has an [overloaded plus](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/plus.html) operation too

save-core/src/commonMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
kgevorkyan: This is `ktoml` function, so I filtered it in such way

Now I created function extension `getTopLevelTomlTables`

save-plugins/fix-and-warn-plugin/src/commonTest/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPluginTest.kt
kgevorkyan: Fixed

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
kgevorkyan: I need to force FixPlugin to work before warnings will be restored into expected file, so I apply `toList`, 
What about warnPlugin - it should work fine without `toList` and was just reinsurance, removed

petertrr: Hmm, but if we simply concatenate two sequences, they will be calculated in the same order, right? Is this a reinsurance for the future, when we will do this in parallel?

kgevorkyan: This changes corresponds not really to the parallel.\
I noticed, that without applying `toList` to fix results, fix plugin will executes only in the end of work progress. (when warnings are already restored)\
However, we need him to execute exactly after when warnings will be removed from expected file and before when they will be restored

petertrr: I see, yeah, I don't see a better solution yet. Please add comments about it, it's not obvious

petertrr (approved):

PR #188, opened 2021-07-21
Integration test

kgevorkyan:
### What's done:
* Create integration test, which executes SAVE binary on the examples


save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        assertTrue(fs.exists(binDir.toPath()))
+
+        val saveExecutableFiles = fs.list(binDir.toPath()).filter { fs.metadata(it).isRegularFile }
+        // Binary should be created at this moment
+        assertTrue(saveExecutableFiles.isNotEmpty())
+
+        val examplesDir = "../examples/kotlin-diktat/"
+
+        val actualSaveBinary = saveExecutableFiles.last()
+        val saveBinName = if (isCurrentOsWindows()) "save.exe" else "save"
```
petertrr, line 47: Isn't it `.kexe` on Linux and mac?

save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        val destination = (examplesDir + saveBinName).toPath()
```
petertrr, line 48:
```suggestion
        val destination = examplesDir.toPath() / saveBinName
```
I think it looks more readable, and you don't need to keep in mind a trailing slash

save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        // Copy latest version of save into examples
+        fs.copy(actualSaveBinary, destination)
+        assertTrue(fs.exists(destination))
+        fs.delete(actualSaveBinary)
```
petertrr, line 53: Why do you delete it?

save-core/build.gradle.kts
```diff
@@ -77,6 +79,17 @@ tasks.withType<KotlinCompile<*>>().forEach {
     it.dependsOn(generateVersionFileTaskProvider)
 }
 
+// Integration test should be able to have access to binary during the execution
+val os = getCurrentOperatingSystem()
+tasks.getByName("jvmTest").dependsOn(
```
petertrr, line 84: Can this task be moved to save-cli project? Currently you get a circular dependency: save-cli depends on save-core for compilation, but save-core depends on save-cli for tests.

petertrr (changes requested):

save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
```diff
@@ -74,10 +74,10 @@ fun createConfigFromArgs(args: Array<String>): SaveProperties {
     val testFiles = configFromCli.testFiles
     if (!testFiles.isNullOrEmpty() && !fs.exists(testFiles[0].toPath())) {
         // FixMe: get(0) to [0] after https://github.com/cqfn/diKTat/issues/1047
-        errorAndExitNotValidDir(testFiles!!.get(0).toPath())
+        errorAndExitNotValidDir(testFiles.get(0).toPath())
     }
     // FixMe: get(0) to [0] after https://github.com/cqfn/diKTat/issues/1047
```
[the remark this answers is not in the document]
kgevorkyan: FixMe is actual, because of opened https://github.com/cqfn/diKTat/pull/1050

I just removed redundant `!!` access, since in these cases we already checked `testFiles` for nullability by `!testFiles.isNullOrEmpty()` and there were warning about it by compiler

save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
kgevorkyan: I need `.exe` on Windows just to show this OS, that current file is executable, on Mac and Linux it doesn't matter in general, it is ELF file, for which we give permissions for execution by `chmod`

save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
kgevorkyan: Thanks, fixed. I thought, that such trick we can apply only when both parts are `okio.path`, however, when right part is `string` it also worked

save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
kgevorkyan: This binary we use just for tests, so if during the development we will usually rerun tests and hold this file I think
`save-cli/bin/..` will contain a lot of garbage

save-core/build.gradle.kts
kgevorkyan: Initially I create this test in `save-cli`, but then I thought, that it is not really good, because in this case in `save-cli` we should also download the `diktat` and `ktlint` and then remove them (by tasks from `save-core`). 

So, since we run tests in parallel, it could lead to some problems: `save-cli` will download resources for it's tests and in the same time `save-core` could delete them, because integration tests from `save-core` could be already finished.

Also I there were some other problems which I didn't remember. However now they could be not actual, and I can try to move this test to `save-cli` again

kgevorkyan: Will wait your opinion about my comment above before changing

save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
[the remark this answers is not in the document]
petertrr: Yes, so you can now change `!!.get(0)` to simply `[0]`, even if the issue is still opened

save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
petertrr: But this is the file, produced by `linkDebugExecutable` task, right? I think it will be very confusing, if it gets deleted during tests. Also, during development they all will have the same name with snapshot version.

save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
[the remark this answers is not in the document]
kgevorkyan: Really, didn't notice

save-core/build.gradle.kts
petertrr: Yeah, that's not a very good structure anyway. I'd suggest to move this test to save-cli and add dependency on `save-cli:jvmTest` to the task, that cleans up downloaded files.

save-core/src/jvmTest/kotlin/org/cqfn/save/core/integration/GeneralTest.kt
kgevorkyan: Yes, but this task we call by ourselves in test. So I don't see any confusing: we create bin file for our needs, and then remove it.

I didn't met situations in SAVE-cli, when I get an snapshot version by `linkDebugExecutable` However, it make sence,
if we will get such situation, then we will remove the only one existing binary.

Removed this line

save-cli/src/jvmTest/kotlin/org/cqfn/save/cli/GeneralTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        val reportFile = examplesDir.toPath() / "save.out.json".toPath()
+        if (fs.exists(reportFile)) {
+            fs.delete(reportFile)
+        }
+
+        val runCmd = if (isCurrentOsWindows()) "" else "sudo chmod +x $saveBinName && ./"
+        val saveFlags = " . --result-output FILE --report-type JSON"
+        // Execute the script from examples
+        val execCmd = "$runCmd$saveBinName $saveFlags"
+        ProcessBuilder(true, fs).exec(execCmd, examplesDir, null)
```
petertrr, line 69: Let's also print stdout of the process, so that it can be viewed in gradle test reports if needed

petertrr (approved):

PR #190, opened 2021-07-22
Precompiled script plugin for kotlin configuration

### LINKED ISSUE #76: Create common gradle configuration for all kotlin targets
At least we need to move common logic like:
masosX64() 
mingwX64()
linuxX64() 

to the common module

petertrr:
### What's done:
* Added kotlin-library.gradle.kts
* Use it in build.gradle.kts
* Refactor some CI-specific gradle properties

Closes #76


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/kotlin-library.gradle.kts
```diff
@@ -0,0 +1,96 @@
...
+kotlin {
+    jvm {
+        compilations.all {
+            kotlinOptions {
+                jvmTarget = "11"
+                freeCompilerArgs = freeCompilerArgs + "-Xjvm-default=all"
+            }
+        }
+    }
+    val nativeTargets = listOf(linuxX64(), mingwX64(), macosX64())
```
kgevorkyan, line 28: Why diktat didn't warn on this line? the first usage of this variable is at 82 line. Something wrong with diktat configuration?

petertrr: Funny thing is, that here we need to declare this variable here to activate targets before disabling some of them in the next `if`. Speaking about diktat, this behavior is intended: 
 https://github.com/cqfn/diKTat/blob/70066f60425d7a092427e783e278fbfa40bebe83/diktat-rules/src/main/kotlin/org/cqfn/diktat/ruleset/rules/chapter3/identifiers/LocalVariablesRule.kt#L204
We are whitelisting several methods, inlcuding `listOf`

PR #193, opened 2021-07-22
Don't run WarnPlugin on Fail testResult in FixAndWarnPlugin

### LINKED ISSUE #192: Don't run WarnPlugin on Fail testResult in FixAndWarnPlugin

Cheshiriks:
What's done:
* Added filter on Fail testResult
Closes #192


save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -88,7 +101,7 @@ class FixAndWarnPlugin(
         // TODO: then warn plugin should look at the fix plugin output for actual warnings, and not execute command one more time.
         // TODO: However it's required changes in warn plugin logic (it's should be able to compare expected and actual warnings from different places),
         // TODO: this probably could be obtained after https://github.com/cqfn/save/issues/164,
-        val warnTestResults = warnPlugin.handleFiles(expectedFiles.map { listOf(it) })
+        val warnTestResults = warnPlugin.handleFiles(expectedFilesWithoutFail.map { listOf(it) })
         return fixTestResults.asSequence() + warnTestResults
```
kgevorkyan, line 105: If I correctly remember, we shouldn't return all results. We need to return only those files, on which fix plugin fails, and warn results on which fix plugin pass

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -70,8 +71,20 @@ class FixAndWarnPlugin(
 
         val fixTestResults = fixPlugin.handleFiles(files).toList()
 
+        val filesAndTheirWarningsMapWithoutFail = filesAndTheirWarningsMap.filter { filesAndTheir ->
```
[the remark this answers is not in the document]
Cheshiriks: Fixed

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPlugin.kt
```diff
@@ -70,6 +71,16 @@ class FixAndWarnPlugin(
 
         val fixTestResults = fixPlugin.handleFiles(files).toList()
 
+        val fixTestResultsFail = fixTestResults.filter { fixResult ->
+            fixResult.status !is Pass
+        }
+
+        val expectedFilesWithPass = expectedFiles.filter { expectedFile ->
+            fixTestResults.filter { fixResult ->
```
kgevorkyan, line 79: You use here almost the same logic for filtering, I think it could be done in more clear way and in one step:

```kotlin
        val (fixTestResultsPassed, fixTestResultsFailed) = fixTestResults.partition { it.status is Pass }
```
Then you will be able to use `fixTestResultsPassed` in `expectedFiles.filter`
and `fixTestResultsFailed` for return statement

Cheshiriks: Yes, thanks :)

kgevorkyan (approved):

PR #198, opened 2021-07-26
Hotfix in JSON reporter

kgevorkyan:
### What's done:
* Comma should be appended every time after first invocation

### DIFF
--- a/save-reporters/src/commonMain/kotlin/org/cqfn/save/reporter/json/JsonReporter.kt
+++ b/save-reporters/src/commonMain/kotlin/org/cqfn/save/reporter/json/JsonReporter.kt
@@ -73,10 +73,11 @@ class JsonReporter(override val out: BufferedSink) : Reporter {
      *
      * @return `true` if comma has been written, `false` otherwise
      */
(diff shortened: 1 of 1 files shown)


save-reporters/src/commonMain/kotlin/org/cqfn/save/reporter/json/JsonReporter.kt
```diff
@@ -73,10 +73,11 @@ class JsonReporter(override val out: BufferedSink) : Reporter {
      *
      * @return `true` if comma has been written, `false` otherwise
```
petertrr, line 74: Please also fix this doc: with current implementation, after the first invocation of this method flag should be switched to false, and we re-assign true somewhere else.

petertrr (approved):

save-reporters/src/commonMain/kotlin/org/cqfn/save/reporter/json/JsonReporter.kt
kgevorkyan: Thanks, fixed

PR #209, opened 2021-08-02
Fix build on Windows in CI

### LINKED ISSUE #207: Fix windows build or disable http client for Windows

petertrr:
### What's done:
* Update build_and_test.yml: gradle-cache-action ignores properties without value, so added explicit `disableRedundantTargets=true`
* Added more logging
* Always close ktor client

Closes #207

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -95,7 +95,7 @@ jobs:
       - name: Cache konan
         uses: actions/cache@v2
         with:
-          path: ${{ env.HOMEPATH }}/.konan
+          path: ${{ env.USERPROFILE }}/.konan
           key: Windows-gradle-konan
           restore-keys: |
             Windows-gradle-konan
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -55,7 +55,7 @@ jobs:
           properties: |
             org.gradle.caching=true
             detekt.multiplatform.disabled=true
-            disableRedundantTargets
+            disableRedundantTargets=true
             enabledExecutables=debug
 
       - name: Upload gradle reports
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -72,32 +72,26 @@ jobs:
...
-      - name: Install curl Windows
-        continue-on-error: true
-        run: pacman -S --noconfirm mingw-w64-x86_64-curl
-        shell: msys2 {0}
-
       - name: Cache konan
         uses: actions/cache@v2
         with:
-          path: ~/.konan
+          path: 'C:\Users\runneradmin\.konan'
```
kgevorkyan, line 94: There is no any env variable for this?

save-core/src/jvmTest/kotlin/org/cqfn/save/core/test/utils/TestUtilsJvm.kt
```diff
@@ -43,10 +43,11 @@ actual suspend fun downloadFile(url: String, fileName: String): String {
     if (!file.exists()) {
         val httpResponse: HttpResponse = client.get(url)
         val responseBody: ByteArray = httpResponse.receive()
+        println("Writing ${responseBody.size} bytes into ${file.path}")
```
kgevorkyan, line 46: May be `logDebug`?

.github/workflows/build_and_test.yml
petertrr: I tried `HOME`, `HOMEPATH` and `USERPROFILE`, they don't seem to work, at least when configuring a step

save-core/src/jvmTest/kotlin/org/cqfn/save/core/test/utils/TestUtilsJvm.kt
petertrr: Yes, changed

kgevorkyan (approved):

PR #216, opened 2021-08-11
Introduce save-common-test

### LINKED ISSUE #208: Delete binaries in integrational tests after all tests are finished
Now I am not deleting `ktlint` and `diktat` after the execution of tests

petertrr:
### What's done:
* Added subproject `save-common-test`
* Cleanup after integration test via gradle task
* Move MockPlugin.kt to the new module
* Added common test utils: `BeforeClass`, `AfterClass` for Native and JVM
* Enable ignored tests in `ClassicWarnTest`
* Move `runTest` into `commonNonJsTest`

Closes #208

Future enhancements:
* We can create `ProcessBuilderMock`, that can be used in unit tests: it does nothing and returns a specified code. Can be passed into plugins in tests.

Caveats:
* Moving `TestReporter` to `save-common-test` is not possible until #5


save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
```diff
@@ -117,6 +118,30 @@ fun FileSystem.readFile(path: Path): String = this.read(path) {
...
 }
 
+/**
+ * Copies [source] with all files and subdirectories into [target]
+ *
+ * @param source a directory to copy
+ * @param target a destination directory
+ */
+fun FileSystem.copyRecursively(source: Path, target: Path) {
+    require(metadata(source).isDirectory)
```
kgevorkyan, line 128: I remember, that we had some problems with `isDirectory` and should be accurate if there weren't updates in `okio`:

Some directories in unix systems can have extended permissions (like for example `/tmp`), which make their type not exactly `directory`, but `sticky directory`.

And `okio` for some reasons doesn't identify them as directory sometimes, however they are directories. 
It's strange, but `/tmp` in Linux and Mac is `sticky directory`, but `metadata(/tmp).isDirectory`  `== true` on Linux, but `== false` on Mac.

Previously for that reason we used `!isRegularFile`

save-common-test/src/commonNonJsMain/kotlin/org/cqfn/save/Utils.kt
```diff
@@ -0,0 +1,34 @@
...
+package org.cqfn.save
+
+import okio.FileSystem
+import okio.Path
+import kotlin.random.Random
+
+/**
+ * Marks method to be executed before all tests, supported on JVM and Native
+ */
+expect annotation class BeforeClass
```
kgevorkyan, line 16: I don't see the usage of these annotations in this PR, is it for future enhancements?

save-common-test/src/commonNonJsMain/kotlin/org/cqfn/save/Utils.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+expect annotation class AfterClass
+
+/**
+ * Create a temporary directory
+ *
+ * @param prefix will be prepended to directory name
+ * @return a [Path] representing the created directory
+ */
+fun FileSystem.createTempDir(prefix: String = "tmp"): Path {
```
akuleshov7, line 29: May be "save-tmp"? As  in this temp-dir should be thousans of directories

save-common-test/src/commonNonJsMain/kotlin/org/cqfn/save/Utils.kt
petertrr: Yeah, I initially wanted to use these annotations, but then implemented resources cleanup as a gradle task. Decided to keep annotations, because they probably will be useful in the future.

save-core/build.gradle.kts
```diff
@@ -85,3 +89,29 @@ tasks.withType<KotlinCompile<*>>().forEach {
     it.dependsOn(generateConfigOptionsTaskProvider)
     it.dependsOn(generateVersionFileTaskProvider)
 }
+
+val ktlintVersion = "0.39.0"
```
akuleshov7, line 93: may be move versions to Versions.kt?

save-core/build.gradle.kts
```diff
@@ -85,3 +89,29 @@ tasks.withType<KotlinCompile<*>>().forEach {
...
+val diktatVersion = "1.0.0-rc.2"
+tasks.register<Download>("downloadTestResources") {
+    src(listOf(
+        "https://github.com/pinterest/ktlint/releases/download/$ktlintVersion/ktlint",
```
akuleshov7, line 97: And these links also to Versions.kt

akuleshov7 (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
petertrr: Thx, I keep forgetting about this. It's strange though, because `!isRegularFile` will be true not only for directories, but also for symlinks, sockets, pipes etc. And here 
https://github.com/square/okio/blob/ffbf472af20f39a18f46365c7a7dcf5a01f84cd1/okio/src/linuxX64Main/kotlin/okio/-LinuxX64PosixVariant.kt#L38
you can see that they check whether the file is a directory in an expected way.

petertrr: Changed, at least for now

save-core/build.gradle.kts
petertrr: Moved

save-common-test/src/commonNonJsMain/kotlin/org/cqfn/save/Utils.kt
petertrr: Renamed

save-core/build.gradle.kts
petertrr: Moved

PR #217, opened 2021-08-18
Updating save to the new Toml API

akuleshov7:
### What's done:
- new API of ktoml (0.2.7)
- changed Int to Long as Toml does not support Int type

### DIFF
--- a/save-core/src/commonNonJsMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
+++ b/save-core/src/commonNonJsMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
@@ -5,6 +5,7 @@
 package org.cqfn.save.core.utils
 
 import org.cqfn.save.core.config.TestConfigSections
+import org.cqfn.save.core.files.readLines
 import org.cqfn.save.core.logging.logError
 import org.cqfn.save.core.plugin.GeneralConfig
 import org.cqfn.save.core.plugin.PluginConfig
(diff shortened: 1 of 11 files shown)


save-core/src/commonNonJsMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
```diff
@@ -14,52 +14,47 @@ import org.cqfn.save.plugins.fix.FixPluginConfig
...
-            )
-            TestConfigSections.GENERAL.name -> this.createPluginConfig<GeneralConfig>(
-                table.fullTableName
-            )
-            else -> throw PluginException(
-                "Received unknown plugin section name in the input: [${table.fullTableName}]." +
-                        " Please check your <$this> config"
-            )
-        }
+    when (table.fullTableName.uppercase().replace("\"", "")) {
```
petertrr, line 29: This means, that deserialized strings still have surrounding quotes?

petertrr (approved):

save-core/src/commonNonJsMain/kotlin/org/cqfn/save/core/utils/TomlUtils.kt
akuleshov7: No, this is only regarding the table name. FullTableName is actually an internal field that should not be used in classic deserialization scenario. 

`a = "abc"` - for key-value pairs of course I do not store quotes.

But we in save want to use these names of tables, so we need to use some technical fields.

PR #219, opened 2021-08-19
Adding the support to Array fields in configs

akuleshov7:
### What's done:
- changing type for tags, excludedTests, includedTests in GeneralConfig
- updating test resources

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
@@ -51,21 +51,21 @@ interface PluginConfig {
  * The logic of the default value processing will be provided in stage of validation
  *
  * @property execCmd a command that will be executed to check resources and emit warnings
- * @property tags FixMe: after ktoml will support lists we should change it
- * @property description
- * @property suiteName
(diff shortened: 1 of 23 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -112,17 +110,17 @@ data class GeneralConfig(
...
             expectedWarningsPattern ?: defaultInputPattern,
         )
     }
 
     private fun errorMsgForRequireCheck(field: String) =
             """
-                    Error: Couldn't find `$field` in [general] section of `$configLocation` config.
-                    Current configuration: ${this.toString().substringAfter("(").substringBefore(")")}
-                    Please provide it in this, or at least in one of the parent configs.
+                        Error: Couldn't find `$field` in [general] section of `$configLocation` config.
```
kgevorkyan, line 121: IDEA's fix again?

kgevorkyan (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
akuleshov7: yes, shit

PR #220, opened 2021-08-30
Create the logic in Plugins, which will manage redirectTo for ProcessBuilder

### LINKED ISSUE #122: Create the logic in Plugins, which will manage `redirectTo` for ProcessBuilder
Now the value of `redirectTo` just hard coded to `null` in Plugins, and it's impossible to configure such option for user

Cheshiriks:
What's done:
* Added logic for redirectTo
Closes #122

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -17,12 +17,14 @@ import okio.Path
  * @property testFiles a list of files (test resources or save.toml configs)
  * @property fs describes the current file system
  * @property useInternalRedirections whether to redirect stdout/stderr for internal purposes
+ * @property redirectTo a file where process output and errors should be redirected. If null, output will be returned as [ExecutionResult.stdout] and [ExecutionResult.stderr].
  */
 abstract class Plugin(
     open val testConfig: TestConfig,
     protected val testFiles: List<String>,
     protected val fs: FileSystem,
-    private val useInternalRedirections: Boolean) {
+    private val useInternalRedirections: Boolean,
+    open val redirectTo: Path?) {
     /**
      * Instance that is capable of executing processes
      */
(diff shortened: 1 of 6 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -118,7 +120,7 @@ class WarnPlugin(
         val execCmd = "${generalConfig.execCmd} ${warnPluginConfig.execFlags} $fileNamesForExecCmd"
 
         val executionResult = try {
-            pb.exec("cd ${testConfig.getRootConfig().location.parent} && $execCmd", null)
```
kgevorkyan, line 121: Same should be done for `FixPlugin`

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -17,12 +17,14 @@ import okio.Path
...
  * @property useInternalRedirections whether to redirect stdout/stderr for internal purposes
+ * @property redirectTo a file where process output and errors should be redirected. If null, output will be returned as [ExecutionResult.stdout] and [ExecutionResult.stderr].
  */
 abstract class Plugin(
     open val testConfig: TestConfig,
     protected val testFiles: List<String>,
     protected val fs: FileSystem,
-    private val useInternalRedirections: Boolean) {
+    private val useInternalRedirections: Boolean,
+    open val redirectTo: Path?) {
```
kgevorkyan, line 27: Why `open`?

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
Cheshiriks: Fixed

kgevorkyan (approved):

PR #228, opened 2021-08-31
Improve logging

### LINKED ISSUE #224: Save logging: need to log something if save have not discover any test files
Now save is not showing anything (need at least to print `Finished` in the end).

petertrr:
### What's done:
* Improve logging

Closes #224


save-core/src/commonNonJsMain/kotlin/org/cqfn/save/core/Save.kt
```diff
@@ -95,6 +92,7 @@ class Save(
             }
         reporter.afterAll()
         reporter.out.close()
+        logInfo("Finished")
```
akuleshov7, line 95: Save processing finished. If the result is unexpected, please use --debug option

akuleshov7 (changes requested):
please check warning msg

akuleshov7 (approved):

PR #229, opened 2021-08-31
Support inline fixers for fix and warn plugin

### LINKED ISSUE #221: Support inline fixers for fix and warn plugin
We need to support the following LIT fixers: 
https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/test/clang-tidy/checkers/abseil-time-comparison.cpp

```
// CHECK-MESSAGES: [[@LINE-1]]:7: warning: perform comparison in the time domain [abseil-time-comparison]
// CHECK-FIXES: absl::FromUnixSeconds(x) <= t1;
```

We need to:
1) Support this logic in a fix and warn plugin
2) To do this we need to generate tmp `Expected` test resource 
3) Compare files and check results

Cheshiriks:
What's done:
* Added support inline fixers for fix and warn plugin
Closes #221


akuleshov7:
Let's close it by for now. As it is not needed

PR #232, opened 2021-09-01
Print stderr from process execution only in debug mode

kgevorkyan:
### What's done:
Print stderr from process execution only in debug mode


save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -102,7 +102,7 @@ class ProcessBuilder(private val useInternalRedirections: Boolean, private val f
         fs.deleteRecursively(tmpDir)
         logDebug("Removed temp directory $tmpDir")
         if (stderr.isNotEmpty()) {
-            logWarn(stderr.joinToString("\n"))
+            logDebug(stderr.joinToString("\n"))
```
akuleshov7, line 105: The following errors occured after executing: <execCmd>

kgevorkyan: Fixed

akuleshov7 (approved):

PR #234, opened 2021-09-01
Improve logging

kgevorkyan:


examples/kotlin-diktat/fix_and_warn/save.toml
```diff
@@ -8,5 +8,5 @@ expectedWarningsPattern = ".*// ;warn:(\\d+):(\\d+) (.*)"
     ["fix and warn".fix]
         execFlags = "-F"
     ["fix and warn".warn]
-        actualWarningsPattern = "(\\w+\\..+):(\\d+):(\\d+): (.*)$"
+        actualWarningsPattern = "(\\w+\\..+):(\\d+):(\\d+): (\\[.*\\].*)$"
```
kgevorkyan, line 11: Examples had several problems, if were executed by hands

examples/kotlin-diktat/warn/save.toml
```diff
@@ -6,6 +6,7 @@ suiteName = "warnings"
 [warn]
 # suffix to detect tests
 testNameSuffix = "Test.kt"
+wildCardInDirectoryMode = "**/*"
```
kgevorkyan, line 9: Same

save-cli/src/nativeMain/kotlin/org/cqfn/save/cli/SaveConfigReader.kt
```diff
@@ -53,12 +54,14 @@ fun SaveProperties.getFields() = this.toString().dropWhile { it != '(' }.drop(1)
 fun createConfigFromArgs(args: Array<String>): SaveProperties {
     // getting configuration from command-line arguments
     val configFromCli = SaveProperties(args)
+    tryToUpdateDebugLevel(configFromCli)
```
kgevorkyan, line 57: Previously we enabled debug only at the `perfomAnalysis` stage

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -165,9 +165,14 @@ class WarnPlugin(
         }.asSequence()
     }
 
-    private fun String.makeThePathRelativeToTestRoot() =
-            this.replace("${testConfig.getRootConfig().directory}", "")
-                .trimStart('/', '\\')
+    private fun String.makeThePathRelativeToTestRoot(): String {
+        val testRoot = testConfig.getRootConfig().directory.toString()
+        if (testRoot == ".") {
```
kgevorkyan, line 170: Bugfix in function in case, when test root dir == '.'

Before, for example for single file `Test1Test.kt` and test root dir == '.' we will get ``Test1Testkt``

However, it's temporary solution, since we already have almost the same function https://github.com/cqfn/save/issues/237

PR #238, opened 2021-09-01
Removing includedTests option and adding functionality for excludedTests

akuleshov7:
### What's done:
- added logic for excludedTests
- updating test resources
- removing includedTests option


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -33,12 +34,19 @@ abstract class Plugin(
...
      */
-    fun execute(): Sequence<TestResult> {
+    fun execute(rootDir: Path, excludedTests: List<String>?): Sequence<TestResult> {
         clean()
-        // todo: pass individual groups of files to handleFiles? Or it will play bad with batch mode?
+
         val testFilesSequence = discoverTestFiles(testConfig.directory)
+            // removing excluded test resources
+            .filterNot {
+                isExcludedTest(it, rootDir, excludedTests)
```
petertrr, line 47: I think it should be moved closer to the place where we filter based on `testFiles` (in `discoverTestFiles`)

petertrr (approved):
LGTM

PR #240, opened 2021-09-02
Move `validateAndSetDefaults` to `Plugin`

petertrr:
### What's done:
* Move `validateAndSetDefaults` from individual plugins to `Plugin`
* Minor refactoring in `WarnPluginConfig`


akuleshov7 (approved):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
```diff
@@ -147,7 +142,7 @@ data class WarnPluginConfig(
             newColumnCaptureGroupOut,
             newMessageCaptureGroupOut,
             exactWarningsMatch ?: true,
-            testName,
```
akuleshov7, line 150: very confusing, that we had testName here...

PR #245, opened 2021-09-03
Dependency updates

petertrr:
### What's done:
* Kotlin 1.5.31
* Diktat 1.0.0-rc.3, code style updates
* Other dependencies: okio, kotlinx.cli, kotlinx.coroutines, junit


petertrr:
There seem to be a problem with newer versions of serialization on macos (reproduced twice in CI builds):
```
GeneralTest[jvm] > examples test()[jvm] FAILED
    kotlinx.serialization.json.internal.JsonDecodingException at GeneralTest.kt:96
There were failing tests
```

PR #255, opened 2021-09-13
Add new parameter for ProcessBuilder : directory

kgevorkyan:
### What's done:
* Add new parameter for ProcessBuilder : directory - where to step over before execution, if necessary
* Extract modifying of `cmd` into separate function

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
@@ -52,8 +52,8 @@ class ProcessBuilder(private val useInternalRedirections: Boolean, private val f
      * Execute [command] and wait for its completion.
      *
      * @param command executable command with arguments
-     * @param redirectTo a file where process output and errors should be redirected. If null, output will be returned as [ExecutionResult.stdout] and [ExecutionResult.stderr].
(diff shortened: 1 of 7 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -62,7 +63,8 @@ class ProcessBuilder(private val useInternalRedirections: Boolean, private val f
         "ReturnCount")
     fun exec(
         command: String,
-        redirectTo: Path?): ExecutionResult {
+        redirectTo: Path?,
+        directory: String? = null): ExecutionResult {
```
akuleshov7, line 67: It's a bad practice to put null as a default

akuleshov7: also, not nullable

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -113,6 +112,28 @@ class ProcessBuilder(private val useInternalRedirections: Boolean, private val f
         return ExecutionResult(status, stdout, stderr)
     }
 
+    private fun modifyCmd(
+        command: String,
+        directory: String?,
```
akuleshov7, line 117: remove nullable here also, code will become more readable

PR #256, opened 2021-09-13
Add extraConfigPattern parameter in fix plugin

### LINKED ISSUE #248: Added extraConfigPattern parameter in fix plugin
Continuation from https://github.com/cqfn/save/pull/236#issuecomment-915828877

Cheshiriks:
What's done:
* Added extraConfigPattern in fix plugin
Closes #248


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
```diff
@@ -163,35 +160,29 @@ data class WarnPluginConfig(
...
     }
 
+    /**
+     * Substitute placeholders in `this.execFlags` with values from provided arguments
+     *
+     * @param extraFlags [ExtraFlags] to be inserted into `execFlags`
+     * @param fileNames file name or names, that need to be inserted into `execFlags`
+     * @return `this.execFlags` with resolved placeholders
+     */
+    fun resolvePlaceholdersFrom(extraFlags: ExtraFlags, fileNames: String): String =
```
petertrr, line 170: Why is this function duplicated in fix and warn plugins? Is there any difference in logic?

petertrr (changes requested):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPluginConfig.kt
Cheshiriks: Fixed

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -158,6 +159,28 @@ abstract class Plugin(
...
 
+    /**
+     * Substitute placeholders in `this.execFlags` with values from provided arguments
+     *
+     * @param execFlags a command that will be executed to check resources and emit warnings
+     * @param extraFlags [ExtraFlags] to be inserted into `execFlags`
+     * @param fileNames file name or names, that need to be inserted into `execFlags`
+     * @return `this.execFlags` with resolved placeholders
+     */
+    fun resolvePlaceholdersFrom(
```
petertrr, line 170: I think this method doesn't use any data from Plugin class. We could make it a top-level function somewhere near ExtraFlagsExtractor or ExtraFlags class declaration, this will also making testing it easier (it won't require creation of Plugin instance).

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -158,6 +159,28 @@ abstract class Plugin(
...
+        execFlags: String?,
+        extraFlags: ExtraFlags,
+        fileNames: String): String =
+            execFlags!!
```
petertrr, line 174: Since we call `!!` anyway, we should either make argument nn-nullable, or use `requireNotNull` with error message here as a precondition

save-plugins/fix-plugin/src/commonNonJsTest/kotlin/org/cqfn/save/plugins/fix/FixPluginTest.kt
```diff
@@ -159,6 +160,62 @@ class FixPluginTest {
         }
     }
 
+    @Test
+    fun `should resolve placeholders`() {
```
petertrr, line 164: And these tests can be moved from plugins into common module, because you are not testing plugin-specific logic here

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
Cheshiriks: Yes. Fixed

PR #257, opened 2021-09-16
TestFiles cli argument should be the list of relative paths

### LINKED ISSUE #235: TestFiles cli argument should be the list of relative paths
For this moment, we support two options:

We can pass to this argument either the relative path to file,

```kotlin
`./save --test-root-path "." warn/chapter1/EnumValueSnakeCaseTest.kt
```
or just the name of it
```kotlin
`./save --test-root-path "." EnumValueSnakeCaseTest.kt
```
Both will give the same result.

However, it can lead to unexpected effects, if user will make a mistake in name in second case
Moreover, for now we have the following check in `discoverTestFiles`
```kotlin
                resourcesGroup.any { path ->
                    testFiles.any { it in path.toString() }
                }
```
thus, even if it will be provided not just the name, but part of path of existing test resource, than we will keep it (unexpected behavior)

**It's suggested to support only the reative paths for `testFiles`**

**Note** On Windows, for execution we need to use extra `\`; 
command should look like:
```kotlin
`./save --test-root-path "." warn\\chapter1\\EnumValueSnakeCaseTest.kt
```
**Information about this should be provided in README**

Cheshiriks:
What's done:
* Added check for the path to the file
Closes #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,11 +141,7 @@ $ save [options] /path/to/tests/Test1 /path/to/tests/Test2
 ```
 SAVE will detect the closest `save.toml` file and use configuration from there.
 
-`Note:` On Windows, for execution we need to use double `\\`.
-Command should look like:
-```bash
-$ save [options] path\\to\\tests\\Test3
-```
+`Note:` On Windows, you may need to use double backslash `\\` as path separator
 
 
 ## <a name="plugins"></a> Using plugins for specific test-scenarios
--- a/README.md
+++ b/README.md
@@ -131,15 +131,21 @@ excludedTests = ["warn/chapter1/GarbageTest.kt", "warn/otherDir/NewTest.kt"], e.
 
 ## Executing specific tests
 It can be useful to execute only a number of tests instead of all tests under a particular `save.toml` config.
-To do so, you want to pass a test file name after all configuration options:
+To do so, you want to pass a relative path to test file after all configuration options:
 ```bash
 $ save [options] /path/to/tests/Test1
 ```
-or a list of file names (separated with spaces)
+or a list of relative paths to test files (separated with spaces)
 ```bash
 $ save [options] /path/to/tests/Test1 /path/to/tests/Test2
 ```
 SAVE will detect the closest `save.toml` file and use configuration from there.
+
+`Note:` On Windows, for execution we need to use double `\\`.
+Command should look like:
+```bash
+$ save [options] path\\to\\tests\\Test3
+```
 
 
 ## <a name="plugins"></a> Using plugins for specific test-scenarios
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -8,7 +8,6 @@ import org.cqfn.save.core.logging.logDebug
 import org.cqfn.save.core.result.TestResult
 import org.cqfn.save.core.utils.ProcessBuilder
 
-import okio.FileNotFoundException
 import okio.FileSystem
 import okio.Path
 import okio.Path.Companion.toPath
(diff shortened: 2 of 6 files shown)


kgevorkyan:
Why this test still worked?

https://github.com/cqfn/save/blob/1a12839d7d3bd9a771997c8c0e3f085ddf2af13c/save-core/src/commonNonJsTest/kotlin/org/cqfn/save/core/PerformAnalysisTest.kt#L25-L33

I think, under this PR we prohibit usage of simple name of test file

`testFiles = listOf("../examples/discovery-test", "MyTest.java")`

Cheshiriks:
There are no success checks in this test. Added a relative path.

save-core/src/commonNonJsTest/kotlin/org/cqfn/save/core/integration/WarnDirTest.kt
```diff
@@ -15,7 +15,7 @@ class WarnDirTest {
 
     @Test
     fun `execute warn plugin on the directory chapter2`() =
-            runTestsWithDiktat(listOf("warn-dir/chapter2"), 1)
+            runTestsWithDiktat(listOf("warn-dir/chapter2/GenericFunctionTest.kt"), 1)
```
kgevorkyan, line 18: Why did you change this?

I suggest to create separate test, if you would like to check the logic with individual test file

README.md
```diff
@@ -131,16 +131,22 @@ excludedTests = ["warn/chapter1/GarbageTest.kt", "warn/otherDir/NewTest.kt"], e.
...
 $ save [options] /path/to/tests/Test1 /path/to/tests/Test2
 ```
 SAVE will detect the closest `save.toml` file and use configuration from there.
 
+`Note:` On Windows, for execution we need to use double `\\`.
+Command should look like:
+```bash
+$ save [options] path\\to\\tests\\Test3
+```
+
```
kgevorkyan, line 149:
```suggestion
`Note:` On Windows, you may need to use double backslash `\\` as path separator
```

save-core/src/commonNonJsTest/kotlin/org/cqfn/save/core/integration/WarnDirTest.kt
Cheshiriks: Forgot to delete this

README.md
Cheshiriks: Fixed

save-plugins/fix-plugin/src/commonNonJsTest/kotlin/org/cqfn/save/plugins/fix/FixPluginTest.kt
```diff
@@ -43,6 +43,22 @@ class FixPluginTest {
         assertEquals("Test1Expected.java", pairs.single().second.name)
     }
 
+    @Test
+    fun `tests should have a relative path`() {
```
kgevorkyan, line 47: Please, try to adapt this test (add cases with relative paths), so that we will able to see, that in case of `relative path` all is fine, but when we give just name of the file, we return nothing

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -84,9 +91,24 @@ abstract class Plugin(
             }
 
         return if (testFiles.isNotEmpty()) {
-            rawTestFiles.filter { rawTestFile ->
-                testFiles.any { it in rawTestFile.test.toString() }
+            val (foundTest, notFoundTest) = rawTestFiles.partition { rawTestFile ->
+                testFiles.any {
+                    try {
+                        if (fs.metadata(it.toPath()).isDirectory) {
```
kgevorkyan, line 97: Find out, that https://github.com/cqfn/save/issues/235 is actual only when `testRootPath == .`, since we connecting the paths (`testRootPath` and path to the test file) in 
https://github.com/cqfn/save/blob/fd397c15412b918a36a01e3271dadff62236bd7a/save-core/src/commonNonJsMain/kotlin/org/cqfn/save/core/Save.kt#L60

Thus, only in case `testRootPath == .` we will obtain incorrect behavior. Result of concatenation will contain only the name of test file.

In other cases, the concatenation 
from `Save.kt` will have incorrect value, which won't pass the condition of
`it in rawTestFile.test.toString()`

So, I think we can simplify the logic and just check, whether the `it.toPath()` exists. And no need to check, whether it is a directory and let the remaining logic as it were before.

save-plugins/fix-plugin/src/commonNonJsTest/kotlin/org/cqfn/save/plugins/fix/FixPluginTest.kt
```diff
@@ -42,7 +42,6 @@ class FixPluginTest {
         assertEquals("Test1Test.java", pairs.single().first.name)
         assertEquals("Test1Expected.java", pairs.single().second.name)
     }
-
```
kgevorkyan, line 45: Please restore this line

kgevorkyan (approved):

PR #258, opened 2021-09-16
Calculate number of tests in PlainTextReporter

### LINKED ISSUE #251: Calculation of a pass-rate
Save is now simply printing test results, but no information about the pass-rate and numbers of passed and failed tests.
Need to print this statistics, that would be useful.

petertrr:
### What's done:
* Added Statistics.kt
* Use it in PlainTextReporter.kt

Closes #251

Now last lines of SAVE's output look like this:
```
| Chapter3 | WarnPlugin | warn-dir\chapter3\GenericFunctionTest.kt | Pass |  |
--------------------------------
SUCCESS: 9 tests, 100% successful, failed: 0, skipped: 0
[INFO]: SAVE has finished execution. You can rerun with --debug for additional information.
```

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/reporter/Statistics.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/reporter/Statistics.kt
@@ -42,12 +42,12 @@ class Statistics {
      */
     @Suppress("WHEN_WITHOUT_ELSE")
     fun updateFrom(event: TestResult) {
-        total.inc()
+        total += 1
         when (event.status) {
-            is Pass -> passed.inc()
-            is Fail -> failed.inc()
-            is Ignored -> skipped.inc()
-            is Crash -> crashed.inc()
+            is Pass -> passed += 1
+            is Fail -> failed += 1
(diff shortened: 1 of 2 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/reporter/Statistics.kt
```diff
@@ -0,0 +1,53 @@
...
+    var crashed: Int = 0
+
+    /**
+     * Update fields of this class based on the received [TestResult]
+     *
+     * @param event an event to update statistics
+     */
+    @Suppress("WHEN_WITHOUT_ELSE")
+    fun updateFrom(event: TestResult) {
+        total += 1
```
kgevorkyan, line 45: What do you think?
```suggestion
        total++
```

kgevorkyan (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/reporter/Statistics.kt
petertrr: Yes, that's better. Changed

PR #259, opened 2021-09-17
Implement reporter of type plain-failed

### LINKED ISSUE #249: Aggregate test results by pass or fail
Now test results are not aggregated. May be we could support some kind of an adaptor that would be able to collect and aggregate test results. For example, in Idea there is a possibility to show only `failed` tests. Could be very useful

petertrr:
### What's done:
* Added PlainOnlyFailedReporter.kt
* Updated docs

Closes #249

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,9 @@ In case you would like to have several plugins to work in your directory with sa
 ```
 
 ## SAVE output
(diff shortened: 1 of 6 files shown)


kgevorkyan (approved):
lgtm

PR #272, opened 2021-09-27
Bugfix: Restore movement of validation method

kgevorkyan:
### What's done:
* Restore movement of validation method from https://github.com/cqfn/save/pull/240
* Fix merge and validation methods: update `configLocation`
* Remove redundant `suiteName`'s without test resources
* Make PLAIN as default value for ReportType

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
@@ -89,7 +89,7 @@ data class GeneralConfig(
             this.excludedTests ?: other.excludedTests,
             this.expectedWarningsPattern ?: other.expectedWarningsPattern,
             this.runConfigPattern ?: other.runConfigPattern,
-        )
+        ).also { it.configLocation = this.configLocation }
     }
 
     override fun validateAndSetDefaults(): GeneralConfig {
@@ -113,7 +113,7 @@ data class GeneralConfig(
             excludedTests ?: emptyList(),
             expectedWarningsPattern ?: defaultExpectedWarningPattern,
             runConfigPattern ?: defaultRunConfigPattern,
-        )
+        ).also { it.configLocation = this.configLocation }
     }
 
     private fun errorMsgForRequireCheck(field: String) =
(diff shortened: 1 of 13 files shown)


petertrr (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -113,7 +113,7 @@ data class GeneralConfig(
             excludedTests ?: emptyList(),
             expectedWarningsPattern ?: defaultExpectedWarningPattern,
             runConfigPattern ?: defaultRunConfigPattern,
-        )
+        ).also { it.configLocation = this.configLocation }
```
kgevorkyan, line 116: Previously, in stage of merging we lose the actual `configLocation` here, and used default value, because of new instance creation

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -89,7 +89,7 @@ data class GeneralConfig(
             this.excludedTests ?: other.excludedTests,
             this.expectedWarningsPattern ?: other.expectedWarningsPattern,
             this.runConfigPattern ?: other.runConfigPattern,
-        )
+        ).also { it.configLocation = this.configLocation }
```
petertrr, line 92: Can we make `configLocation` a constructor parameter, so that we don't forget to add `also` in the future? And generally everything would look more consistent

kgevorkyan: I think the main thing, why we initially took out it from ctor, because of serialization and ability to provide it in `toml` files - but this field is used only for internal purposes and shouldn't be available from outside.

On the other hand - it's really easy to forget about updating of this field in case of creation of a new instance, like we did. However, the only `*PluginConfig.kt` classes for now are responsible for configs creation, and it seems, that we will not expand this classes a lot in the near future.

I would better leave this field as is, at least for now.

petertrr: Ok, let's leave it as is

PR #276, opened 2021-09-30
Adding a small workaround in logging: --quiet option

akuleshov7:


save-core/src/commonNonJsTest/kotlin/org/cqfn/save/core/integration/ClassicWarnTest.kt
```diff
@@ -41,6 +41,17 @@ class ClassicWarnTest {
                 ), 3
             )
 
+    @Test
+    @Ignore
```
akuleshov7, line 45: this test will be fixed in 277

buildSrc/src/main/resources/config-options.json
```diff
@@ -28,7 +28,7 @@
       "kotlinType": "kotlin.Boolean",
       "fullName" : "quiet",
       "shortName" : "q",
-      "description" : "Do not log anything",
+      "description" : "Log only test results without any warnings",
```
petertrr, line 31:
```suggestion
      "description" : "Log only test results (if `--result-output` is `stdout`) without any warnings",
```

OptionsTable.md
```diff
@@ -6,7 +6,7 @@ Most (except for `-h` and `-prop`) of the options below can be passed to a SAVE
 | parallel | parallel-mode | Whether to enable parallel mode | false |
 | t | threads | Number of threads | 1 |
 | d | debug | Turn on debug logging | false |
-| q | quiet | Do not log anything | false |
+| q | quiet | Log only test results without any warnings | false |
```
petertrr, line 9: Will errors still be logged? It's not clear from the description

OptionsTable.md
```diff
@@ -6,7 +6,7 @@ Most (except for `-h` and `-prop`) of the options below can be passed to a SAVE
 | parallel | parallel-mode | Whether to enable parallel mode | false |
 | t | threads | Number of threads | 1 |
 | d | debug | Turn on debug logging | false |
-| q | quiet | Do not log anything | false |
+| q | quiet | Log only test results (if `--result-output` is `stdout`) without any warnings (only with critical errors) | false |
```
petertrr, line 9: I think it's different now, from what's in json

petertrr (approved):

OptionsTable.md
akuleshov7: updated, thanks

PR #279, opened 2021-10-01
Incorrect directory matching

### LINKED ISSUE #277: Incorrect directory matching in case of same directory name in the directory hierarchy
Continuation of: https://github.com/cqfn/save/pull/276/files#r719567913

Cheshiriks:
What's done:
* Fixed incorrect directory matching in case of same directory name in the directory hierarchy
Closes #277

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -4,7 +4,7 @@ import org.cqfn.save.core.config.TestConfig
 import org.cqfn.save.core.config.isSaveTomlConfig
 import org.cqfn.save.core.files.createRelativePathToTheRoot
 import org.cqfn.save.core.files.findDescendantDirectoriesBy
-import org.cqfn.save.core.files.parents
+import org.cqfn.save.core.files.parentsWithSelf
 import org.cqfn.save.core.logging.logDebug
 import org.cqfn.save.core.result.TestResult
 import org.cqfn.save.core.utils.ProcessBuilder
@@ -110,8 +110,6 @@ abstract class Plugin(
             rawTestFiles
         }
     }
-
-    private fun Path.parentsWithSelf() = listOf(this) + this.parents().toList()
 
     private fun isExcludedTest(testFiles: TestFiles, excludedTests: List<String>?): Boolean {
         // common root of the test repository (not a location of a current test)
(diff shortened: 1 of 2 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -91,7 +92,11 @@ abstract class Plugin(
             val foundTests = rawTestFiles.filter { rawTestFile ->
                 testFiles.any {
                     if (fs.exists(it.toPath())) {
-                        it in rawTestFile.test.toString()
+                        (it.toPath().parents().toList() + listOf(it.toPath())).all { testFile ->
```
petertrr, line 95: Pls add method `parentsWithSelf` or a parameter `parents(withSelf: Boolean = false)`

petertrr, line 95:
```suggestion
                        (it.toPath().parents().toList() + listOf(it.toPath())).all { part ->
```
This argument name makes no sense, it implies that you are traversing `testFiles`, but you are not

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -91,7 +92,11 @@ abstract class Plugin(
...
+                            (rawTestFile.test.parents().toList() + listOf(rawTestFile.test)).any { rawTestFileDir ->
+                                testFile.toString() == rawTestFileDir.toString()
```
petertrr, line 97: Is this the logic we are aiming for, when they have at least one common parent directory?

petertrr (changes requested):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
Cheshiriks: Not really. All parent directories of testFiles submitted by the user must match at least one parent directorie rawTestFiles

petertrr: We can compare parents of `rawTestFile` with entire `testFile`

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
Cheshiriks: Fixed

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
Cheshiriks: Fixed

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -113,6 +111,8 @@ abstract class Plugin(
         }
     }
 
+    private fun Path.parentsWithSelf() = listOf(this) + this.parents().toList()
```
petertrr, line 114: Pls move to FileUtils

petertrr (approved):

PR #283, opened 2021-10-04
Fix flaky integration test

kgevorkyan:
### What's done:
* Enable integration test and update it according latest changes in SAVE


kgevorkyan:
Seems main problem are gone after https://github.com/diktat-static-analysis/save/pull/310

From the 6 CI executions there was only one failure on MacOS, on other OS all worked fine every execution

save-cli/src/jvmTest/kotlin/org/cqfn/save/cli/GeneralTest.kt
```diff
@@ -27,9 +32,16 @@ import kotlinx.serialization.json.Json
 @OptIn(ExperimentalSerializationApi::class)
 class GeneralTest {
     private val fs = FileSystem.SYSTEM
+    private val json = Json {
```
petertrr, line 35: Can we reuse this `Json` from somewhere else? We could have `save-reporters` as a test dependency here (IIRC, there is the same polymorphic serializer there?)

petertrr:
> Seems main problem are gone after #310
> 
> From the 6 CI executions there was only one failure on MacOS, on other OS all worked fine every execution

Were the errors that you faced same as in #311 ?

save-cli/src/jvmTest/kotlin/org/cqfn/save/cli/GeneralTest.kt
kgevorkyan: Refactored

kgevorkyan:
> > Seems main problem are gone after #310
> > From the 6 CI executions there was only one failure on MacOS, on other OS all worked fine every execution
> 
> Were the errors that you faced same as in #311 ?

Actually not, the exception were like `test resources were not found`, as if test files were removed
At the same moment several tests for plugins were failed too

The reason for now is unclear and such situation were not reproduced

petertrr (approved):

PR #288, opened 2021-10-11
Timeout: save should stop and fail tests if the timeout is reached

### LINKED ISSUE #274: Timeout: save should stop and fail tests if the timeout is reached
We need to add a configurable timeout to save tests. If test runs more than N seconds, test should be stopped and marked as failed

Cheshiriks:
What's done:
* Added timeout for tests
Closes #274


petertrr (approved):

save-core/src/commonNonJsTest/kotlin/org/cqfn/save/core/integration/ClassicWarnTest.kt
```diff
@@ -82,13 +84,13 @@ class ClassicWarnTest {
         )
     }
 
-    @Test
```
kgevorkyan: You removed annotation here

PR #307, opened 2021-11-05
ignoreLines feature for filtering the input file

### LINKED ISSUE #133: Support ignoreLines for all plugins: warn, fix, fix and warn

### LINKED ISSUE #221: Support inline fixers for fix and warn plugin

sanyavertolet:
feature/ignoreLines

This pull request closes #133 

What's done:
* [ ] Added ignoreLines option for Warn plugin.
* [ ] Added tests for ignoreLines and Warn plugin.
* [ ] Added ignoreLines option for Fix plugin.
* [ ] Added tests for ignoreLines and Fix plugin.
* [ ] Implemented copying a file in Warn plugin in order to filter out lines that match any regex from ignoreLines.


examples/kotlin-diktat/fix/smoke/src/main/kotlin/org/cqfn/save/IgnoreLinesTest/IgnoreLines/save.toml
```diff
@@ -0,0 +1,7 @@
+[general]
+    tags = ["FixIgnoreLines"]
+    description = "Test for ignoreLines option in Fix"
+    suiteName = "Autofix: Smoke Tests"
+
+[fix]
+    ignoreLines = ["// IGNORE_ME"]
```
akuleshov7, line 7: may be test with regex here?

akuleshov7: and better to test with several different regexes

sanyavertolet: Done

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -74,6 +79,7 @@ data class GeneralConfig(
     val runConfigPattern: Regex? = null,
 ) : PluginConfig {
     override val type = TestConfigSections.GENERAL
+    override var ignoreLinesPatterns: MutableList<Regex> = mutableListOf()
```
akuleshov7, line 82: why `var` here?

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
```diff
@@ -25,6 +25,11 @@ interface PluginConfig {
      */
     val type: TestConfigSections
 
+    /**
+     * list of regexes to be ignored
+     */
+    var ignoreLinesPatterns: MutableList<Regex>
```
akuleshov7, line 31: why `var` here?

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPluginConfig.kt
```diff
@@ -26,14 +26,19 @@ data class FixAndWarnPluginConfig(
     override var configLocation: Path = "undefined_toml_location".toPath()
     override val resourceNamePatternStr: String = "(${fix.resourceNamePatternStr})|(${warn.resourceNamePatternStr})"
 
+    @Transient
```
akuleshov7, line 29: why `@Transient`? I though that we deserialize this value from toml config

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPluginConfig.kt
```diff
@@ -49,6 +54,11 @@ data class FixAndWarnPluginConfig(
         return FixAndWarnPluginConfig(
             fix.validateAndSetDefaults(),
             warn.validateAndSetDefaults()
-        ).also { it.configLocation = this.configLocation }
+        ).also {
+            it.configLocation = this.configLocation
+        }
+    }
+    companion object {
+        internal val defaultIgnoreLines: MutableList<Regex> = mutableListOf()
```
akuleshov7, line 62: Do we really need an extra constant for it? May be you will simply set the default value (`mutableListOf()`) in place, were it is created?

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -128,20 +130,32 @@ class FixPlugin(
...
-    private fun createTestFile(path: Path, generalConfig: GeneralConfig): Path {
+    private fun createTestFile(
+        path: Path,
+        generalConfig: GeneralConfig,
+        fixPluginConfig: FixPluginConfig): Path {
         val pathCopy: Path = constructPathForCopyOfTestFile(FixPlugin::class.simpleName!!, path)
         createTempDir(pathCopy.parent!!)
 
-        val expectedWarningPattern = generalConfig.expectedWarningsPattern
+        val defaultIgnoreLinesPatterns: MutableList<Regex> = mutableListOf()
```
akuleshov7, line 140: why is it named as `defaultIgnoreLinesPatterns`? Why "default"? You are adding values to this list

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -128,20 +130,32 @@ class FixPlugin(
...
         fs.write(fs.createFile(pathCopy)) {
-            fs.readLines(path).forEach {
-                if (expectedWarningPattern == null || !generalConfig.expectedWarningsPattern!!.matches(it)) {
+            fs.readLines(path)
+                .filter { line ->
+                    fixPluginConfig.ignoreLines?.let {
+                        fixPluginConfig.ignoreLinesPatterns.none { pattern -> pattern.matches(line) }
+                    }
+                        ?: run {
+                            defaultIgnoreLinesPatterns.none {regex -> regex.matches(line) }
```
akuleshov7, line 151: diktat is not working? where are spaces after `{`

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -128,20 +130,32 @@ class FixPlugin(
...
+        generalConfig.expectedWarningsPattern?.let { defaultIgnoreLinesPatterns.add(it) }
+        generalConfig.runConfigPattern?.let { defaultIgnoreLinesPatterns.add(it) }
 
         fs.write(fs.createFile(pathCopy)) {
-            fs.readLines(path).forEach {
-                if (expectedWarningPattern == null || !generalConfig.expectedWarningsPattern!!.matches(it)) {
+            fs.readLines(path)
+                .filter { line ->
+                    fixPluginConfig.ignoreLines?.let {
+                        fixPluginConfig.ignoreLinesPatterns.none { pattern -> pattern.matches(line) }
```
akuleshov7, line 148: use `it` in such small lambdas, it looks better with such syntax sugar

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -128,20 +130,32 @@ class FixPlugin(
...
+        val defaultIgnoreLinesPatterns: MutableList<Regex> = mutableListOf()
+        generalConfig.expectedWarningsPattern?.let { defaultIgnoreLinesPatterns.add(it) }
+        generalConfig.runConfigPattern?.let { defaultIgnoreLinesPatterns.add(it) }
 
         fs.write(fs.createFile(pathCopy)) {
-            fs.readLines(path).forEach {
-                if (expectedWarningPattern == null || !generalConfig.expectedWarningsPattern!!.matches(it)) {
+            fs.readLines(path)
+                .filter { line ->
+                    fixPluginConfig.ignoreLines?.let {
```
akuleshov7, line 147: what will happen here if `ignoreLines == null`? Sorry, if I am not completely understand this logic

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -128,20 +130,32 @@ class FixPlugin(
...
         createTempDir(pathCopy.parent!!)
 
-        val expectedWarningPattern = generalConfig.expectedWarningsPattern
+        val defaultIgnoreLinesPatterns: MutableList<Regex> = mutableListOf()
+        generalConfig.expectedWarningsPattern?.let { defaultIgnoreLinesPatterns.add(it) }
+        generalConfig.runConfigPattern?.let { defaultIgnoreLinesPatterns.add(it) }
 
         fs.write(fs.createFile(pathCopy)) {
-            fs.readLines(path).forEach {
-                if (expectedWarningPattern == null || !generalConfig.expectedWarningsPattern!!.matches(it)) {
```
akuleshov7, line 139: `warningsPattern` will be now included into the `ignoreLinesPatterns`?

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
```diff
@@ -61,14 +66,26 @@ data class FixPluginConfig(
             this.batchSeparator ?: other.batchSeparator,
             this.resourceNameTestSuffix ?: other.resourceNameTestSuffix,
             this.resourceNameExpectedSuffix ?: other.resourceNameExpectedSuffix,
-        ).also { it.configLocation = this.configLocation }
+            other.ignoreLines?.let {
+                this.ignoreLines?.let { other.ignoreLines.union(this.ignoreLines) } ?: other.ignoreLines
+            }?.toMutableList() ?: this.ignoreLines
```
akuleshov7, line 71: why do you need to convert it to `mutableList`? I thought that you created this field as a mutableList from a very beginning.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
```diff
@@ -61,14 +66,26 @@ data class FixPluginConfig(
...
+        ).also {
+            it.configLocation = this.configLocation
+        }
     }
 
+    // due to probable bug in ktoml, ignoreLines = [] and no ignoreLines is ktoml are parsed to be mutableListOf("null")
```
akuleshov7, line 77: https://github.com/akuleshov7/ktoml/issues/77

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
```diff
@@ -61,14 +66,26 @@ data class FixPluginConfig(
...
         resourceNameTest,
-        resourceNameExpected
-    ).also { it.configLocation = this.configLocation }
+        resourceNameExpected,
+        ignoreLines?.filter { it != "null" }?.toMutableList() ?: ignoreLines
+    ).also {
+        it.configLocation = this.configLocation
+    }
+    companion object {
+        internal val defaultIgnoreLines: MutableList<Regex> = mutableListOf()
```
akuleshov7, line 89: same as above

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
akuleshov7: gradlew.bat diktatFixAll

akuleshov7 (changes requested):
please see comments above

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
sanyavertolet: `val` should be here

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/PluginConfig.kt
sanyavertolet: `val` should be here

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPluginConfig.kt
sanyavertolet: The point it that we deserialize `ignoreLines` from toml config. `ignoreLinesPatterns` is initialized with `ignoreLines` that we parse from config so it is not deserialized directly from toml.

save-plugins/fix-and-warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPluginConfig.kt
sanyavertolet: To my mind it looked nicer. But ok, it seems that we don't need it here.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
sanyavertolet: Interesting. Seems that this case is ignored by diktat.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
sanyavertolet: If ignoreLines is not set in config file, I should use `expectedWarningsPattern` and `runConfigPattern` by default. So the values I add to this list are values for default case. In this way I think default is ok here.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
sanyavertolet: I have two variables: `ignoreLines` and `ignoreLinesPatterns`. We deserialize `ignoreLines` from toml config so it is nullable. Than we use `ignoreLinesPatterns` is a mutable list of Regex and it is not nullable.
In fix plugin there are three cases: `ignoreLines` is empty, `ignoreLines` is null or `ignoreLines` is a list of regular expressions. Here I check if `ignoreLines` is null

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
sanyavertolet: User should explicitly put `warningPattern` into `ignoreLines` in toml config.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPluginConfig.kt
sanyavertolet: The point is that we use .union(), which returns Set.

akuleshov7 (approved):

akuleshov7:
ok, let's test it

PR #308, opened 2021-11-08
Update to kotlin 1.6.0, okio 3.0.0, ktoml 0.2.9

petertrr:
### What's done:
* Update Kotlin 1.5.31 -> 1.6.0
* Update okio 3.0.0-alpha.11 -> 3.0.0
* Update ktoml 0.2.6 -> 0.2.9 (ktoml-file built with okio-3.0.0)
* Remove `optIn`s


akuleshov7 (approved):
Mostly done with an upgrade of ktoml, please include the updated ktoml into this PR after it's done

akuleshov7:
`readUtf8Line` and `System` are not working in okio3

petertrr:
> FYI: `FileSystem.System` is not working in okio3 in common code:

Yes, that's why we have `commonNonJs` source set in save. They've changed it in one of the early alphas already. Common code should use the interface, and caller should inject actual FS implementation.

petertrr:
@akuleshov7 actually we'll have to update Kotlin in this PR too, because build fails with incompatible ABI now.

akuleshov7 (approved):

Issue #156, opened 2021-06-29
Remove tmp directories, created by plugins after their execution


kgevorkyan:
It seems, that we remove tmp directories only before plugin execution by `clean()` method. 
It's a good point, because if SAVE crashed on something during the previous execution, we anyway will remove old trash.

However, we hold all temp files after SAVE's work, it's seems to be not a good style, and probably we should remove
`tmp files` before **and** after execution

kgevorkyan:
Under this task it also should be refactored `cleanupTempDir` it's abstract method in `Plugin`, which have the same logic in
`Warn` and `Fix`, instead of `name` of directories. This name could be a property of  corresponding plugins, and logic could be moved in to common logic

petertrr:
We also should reconsider the way temporary directories are named. Now we usually use Plugin's `::class.simpleName`, but this causes two parallel executions to clash. Now we are facing it only as flaking tests (e.g. I just saw on CI `java.io.FileNotFoundException: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/FixPlugin/fix_and_warn/smoke/src/main/kotlin/org/cqfn/save/chapter1/Bug1Test.kt (No such file or directory)`, while native test is executed correctly. I believe this is caused by clash of directory names, because similar issues have been fixed, when we started using random directory names for our tests.

We should probably create a root tmp dir (like `"$FileSystem.SYSTEM_TEMPORARY_DIRECTORY/save-tool"`) and delete it only after execution. And we can encapsulate all this logic in some sort of common tmp directory provider, which can also track a list of created directories.

petertrr:
See also https://github.com/diktat-static-analysis/save/pull/310#discussion_r749432400

PR #310, opened 2021-11-15
Random directory name in `FixPlugin` for temporary files, fixes for IO

petertrr:
### What's done:
 * Use random directory name in FixPlugin for temporary files
 * Use `Files.delete` instead of `File#delete` for `deleteRecursively` on JVM (motivation, for example, [here](http://mail.openjdk.java.net/pipermail/core-libs-dev/2012-August/011314.html))
 * Close inputStream in `ProcessBuilder` by using `useLines`
 * Remove workaround with platform target tests ordering
 * Remove redundant operations from workflow (curl is no longer used)
 * Increase timeout in examples to make tests more stable

### TODO (for separate PRs):
* Sometimes integration tests mutate files in examples (at least, when executed locally), including line separators - does that mean, that our logic mutates sources? (#312)
* Improve working with temporary directories, put back tests for `FixPlugin`'s temporary files (#156)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -21,18 +21,6 @@ jobs:
         with:
           java-version: 1.11
 
-      - name: Install curl Linux
-        continue-on-error: true
-        if: ${{ runner.os == 'Linux' }}
-        run: sudo apt-get update && sudo apt-get install -y libcurl4-openssl-dev
-        shell: bash
-
-      - name: Install curl Mac
-        continue-on-error: true
-        if: ${{ runner.os == 'macOS' }}
-        run: brew install curl
-        shell: bash
-
       - name: Cache konan
         uses: actions/cache@v2
         with:
@@ -74,11 +62,6 @@ jobs:
     name: Build and test on Win
     runs-on: windows-latest
     steps:
-      - uses: msys2/setup-msys2@v2
-        with:
-          msystem: MINGW64
-          update: true
-          install: git mingw-w64-x86_64-toolchain mingw-w64-x86_64-curl
       - uses: actions/checkout@v2
       - name: Set up JDK 11
         uses: actions/setup-java@v1
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
@@ -98,10 +98,7 @@ fun FileSystem.createFile(pathString: String): Path = createFile(pathString.toPa
  * @param path path to a new file
  * @return [path]
  */
-fun FileSystem.createFile(path: Path): Path {
-    sink(path).close()
-    return path
-}
+expect fun FileSystem.createFile(path: Path): Path
 
 /**
  * @param path a path to a file
@@ -210,6 +207,8 @@ fun Path.getCurrentDirectory() = if (fs.metadata(this).isRegularFile) {
  * @return a list of parent directories including itself
  */
 fun Path.parentsWithSelf() = listOf(this) + this.parents().toList()
+
+expect fun FileSystem.myDeleteRecursively(path: Path)
 
 /**
  * Create relative path from the current path to the root
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
@@ -98,7 +98,7 @@ fun FileSystem.createFile(pathString: String): Path = createFile(pathString.toPa
  * @param path path to a new file
  * @return [path]
  */
-expect fun FileSystem.createFile(path: Path): Path
+expect fun FileSystem.createFile(path: Path, overwrite: Boolean = true): Path
 
 /**
  * @param path a path to a file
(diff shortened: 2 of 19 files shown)


petertrr:
By this time CI build has succeeded 3/3 on Linux and Win and 2/3 times on MacOS (the only fail is because of test with timeout, and it's most probably a separate issue).

save-plugins/fix-and-warn-plugin/src/commonNonJsTest/kotlin/org/cqfn/save/plugins/fixandwarn/FixAndWarnPluginTest.kt
```diff
@@ -91,17 +88,12 @@ class FixAndWarnPluginTest {
...
+        println("Results $results")
         assertEquals(1, results.count(), "Size of results should equal number of pairs")
-        // Check FixPlugin results
-        val tmpDir = (FileSystem.SYSTEM_TEMPORARY_DIRECTORY / FixPlugin::class.simpleName!!)
-        assertTrue("Files should be identical") {
-            // Additionally ignore warnings in expected file
-            diff(fs.readLines(tmpDir / "Test1Test.java"), fs.readLines(expectedFile).filterNot { it.contains("warn") })
-                .deltas.isEmpty()
-        }
+        // FixMe: Check FixPlugin results (assert that temporary file with formatted code has been created)
```
petertrr, line 96: I imagine, we should extract logic of temporary directory creation into a service-like component. That component can have a separate implementation for tests, when we know exactly which temporary directory has been created. For now, relying on hard-coded names makes logic too inflexible and leads to bugs in concurrent environment.

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -141,7 +142,10 @@ class FixPlugin(
             }
 
     private fun createTestFile(path: Path, generalConfig: GeneralConfig): Path {
-        val pathCopy: Path = constructPathForCopyOfTestFile(FixPlugin::class.simpleName!!, path)
+        val pathCopy: Path = constructPathForCopyOfTestFile(
+            "${FixPlugin::class.simpleName!!}-${Random.nextInt()}",
```
kgevorkyan, line 146: Now we can't retrieve this random value in `cleanupTempDir()`, may be at least add some fixme?

.github/workflows/build_and_test.yml
```diff
@@ -21,18 +21,6 @@ jobs:
         with:
           java-version: 1.11
 
-      - name: Install curl Linux
```
kgevorkyan, line 24: Why we don't need this anymore?

save-common/src/commonMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
```diff
@@ -211,6 +211,13 @@ fun Path.getCurrentDirectory() = if (fs.metadata(this).isRegularFile) {
  */
 fun Path.parentsWithSelf() = listOf(this) + this.parents().toList()
 
+/**
+ * Delete this directory and all other files and directories in it
+ *
+ * @param path a path to a directory
+ */
+expect fun FileSystem.myDeleteRecursively(path: Path)
```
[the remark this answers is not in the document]
petertrr: Main problem - they use `File#delete`, which doesn't give any reasons why deletion has failed. Also, I thought this method is temporary and we should propose using `Files.delete` into okio itself. I've added a link, why `Files.delete` is better than `File#delete` in PR's description.

.github/workflows/build_and_test.yml
petertrr: This was needed, when we were trying to download test resoruces (`jar`s) in test themselves, and now we are doing it via gradle task

save-common/src/jvmMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
```diff
@@ -6,6 +6,23 @@
...
 
 actual val fs: FileSystem = FileSystem.SYSTEM
+
+/**
+ * Delete this directory and all other files and directories in it
+ *
+ * @param path a path to a directory
+ */
+actual fun FileSystem.myDeleteRecursively(path: Path) {
+    path.toFile().walkBottomUp().forEach {
```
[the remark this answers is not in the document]
petertrr: Because `Files.delete` expects a file or an empty directory, so we should cleanup directory contents first

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
petertrr: Ok, I'll also link #156 to these `fixme`s

kgevorkyan (approved):

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
petertrr: I added (temporarily) this directory as a mutable class property. This way it can be properly cleaned up

PR #315, opened 2021-11-19
Feature support functionality for ignoring warning messages

kgevorkyan:
### What's done:
* Support functionality for ignoring warning messages
* Add test


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/utils/Warning.kt
```diff
@@ -33,19 +33,22 @@ data class Warning(
...
                                    line: Int?,
                                    columnGroupIdx: Long?,
                                    messageGroupIdx: Long,
+                                   benchmarkMode: Boolean,
 ): Warning? {
     val groups = warningRegex.find(this)?.groups ?: return null
 
     val column = getRegexGroupSafe(columnGroupIdx, groups, this, "column number")?.toIntOrNull()
-    val message = getRegexGroupSafe(messageGroupIdx, groups, this, "warning message")!!.trim()
+    val message = if (!benchmarkMode) getRegexGroupSafe(messageGroupIdx, groups, this, "warning message")!!.trim() else " "
```
petertrr, line 51: Why `" "` and not `""`?

petertrr (approved):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/utils/Warning.kt
kgevorkyan: We have several unchecked cases in `StringUtils`, e.g. we don't consider situations, when we've got an empty strings and it leads to the exceptions

This PR was aimed to be as fast solution, so I didn't fixed `StringUtils` here and used this workaround

akuleshov7 (approved):

akuleshov7:
The only problem I see right now in Juliet that they in some cases can write a warning in test resource and after that have SEVERAL blank lines before the real code line that this warning points to

PR #324, opened 2021-11-29
Updating ktoml from 0.2.7 to 0.2.8 after the release

akuleshov7:
### What's done:
- ktoml 0.2.8 has the logic with nullable arrays


petertrr:
I think we should get all updates (Kotlin to 1.6, okio to 3.0.0, ktoml to 0.2.8) in one PR (e.g. #308) to avoid all incompatibilities. Or at least update Kotlin and Ktoml in the same PR

Issue #333, opened 2021-12-12
Failing test on MacOS


petertrr:
`org.cqfn.save.core.integration.ClassicWarnTest` fails on MacOS (probably, after update to Kotlin 1.6.0 and other related dependency updates) with `ProcessExecutionException: Stream closed` being thrown instead of `ProcessTimeoutException`

akuleshov7:
Oh :((

petertrr:
I'll temporary disable this test, it makes CI flaky and unstable again :(

kgevorkyan:
Reproduce it again in `executing warn plugin on save-toml file in directory` test at https://github.com/analysis-dev/save/actions/runs/1859663048

PR #340, opened 2021-12-15
Fix performance regression in test resources detection

### LINKED ISSUE #339: Performance: save-cli is extremely slow with long `testFiles` parameter
When repo with tests has a lot of tests and test suites (~40k) and `testFiles` is long (even about 10), save-cli takes huge time to start operating. For `testFiles.size == 5` it takes about 90 seconds to start processing first real save.toml

petertrr:
### What's done:
* Check `includedSuites`/`excludedSuites` before checking plugin resources
* Convert sequence to list in`discoverTestFiles` to make checking of missing `testFiles` more performant
* Reduce number of `fx.exists` calls in `Plugin.kt`
* More trace logging
* Disable a test due to #333 
* Revert version of `actions/upload-artifact` to v2, because the bug is fixed

Closes #339

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
@@ -92,16 +92,21 @@ abstract class Plugin(
     @Suppress(
         "TOO_LONG_FUNCTION",
         "TOO_MANY_LINES_IN_LAMBDA",
+        "AVOID_NULL_CHECKS",
         "SwallowedException",
     )
     fun discoverTestFiles(root: Path): Sequence<TestFiles> {
         val rawTestFiles = rawDiscoverTestFiles(root.resourceDirectories())
             .filterNot { fs.metadata(it.test).isDirectory }
 
+        // fixme: move this filtering on higher level under https://github.com/diktat-static-analysis/save/issues/336
+        val existingTestFiles = testFiles.map {
+            it.takeIf { fs.exists(it.toPath()) }
+        }
         return if (testFiles.isNotEmpty()) {
             val foundTests = rawTestFiles.filter { rawTestFile ->
-                testFiles.any { testFile ->
-                    if (fs.exists(testFile.toPath())) {
+                existingTestFiles.any { testFile ->
+                    if (testFile != null) {
                         (rawTestFile.test.parentsWithSelf()).any { rawTestFileDir ->
                             testFile == rawTestFileDir.toString()
                         }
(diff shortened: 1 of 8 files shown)


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -202,14 +207,14 @@ abstract class Plugin(
...
-     * This takes into account, that if underlying directory contains it's own SAVE config,
+     * This takes into account, that if underlying directory contains its own SAVE config,
      * then this plugin shouldn't touch these resources; it should be done by plugins from that config.
      *
      * @return a sequence of directories possibly containing this plugin's test resources
      */
     private fun Path.resourceDirectories(): Sequence<Path> = findDescendantDirectoriesBy(true) { file ->
         // this matches directories which contain their own SAVE config
-        fs.metadata(file).isRegularFile || fs.list(file).none { it.isSaveTomlConfig() }
+        fs.metadata(file).isDirectory && fs.list(file).none { it.isSaveTomlConfig() }
```
kgevorkyan, line 217: how it worked before? that's a different logic now

kgevorkyan:
Do you have some comparison results by these improvements?

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
petertrr: It was working, because inside `findDescendantDirectoriesBy` we filter only directories, as the name of the method assumes. So the predicate was passing, but on the very next step the file was filtered out. So I changed the predicate to make it more consistent.

petertrr:
> Do you have some comparison results by these improvements?

Yes, overall it used to take ~1 minute only to merge the top level config and similar times for other configs (as in logs in the issue), and now it takes several seconds:
```
[DEBUG] 2021-12-15 13:50:33: Start merging configs for src/testcases/save.toml
[DEBUG] 2021-12-15 13:50:36: The following tests were not found: 67. Try to make sure you have specified the correct relative path to the files.
[DEBUG] 2021-12-15 13:50:36: Plugin <WarnPlugin> in config file src/testcases/save.toml has no test resources; it's config will only be used for inheritance
[DEBUG] 2021-12-15 13:50:36: Start merging configs for src/testcases/CWE111_Unsafe_JNI/save.toml
```
Overall, the whole execution with batch of 25 tests takes just a couple of minutes (excluding actual tool invocation).

kgevorkyan (approved):

PR #341, opened 2021-12-15
Fix of incorrect paths of test files in FixPlugin results

kgevorkyan:
### What's done:
* Restore `relativeTo` from `okio` to our realization

### DIFF
--- a/save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
+++ b/save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
@@ -2,6 +2,7 @@ package org.cqfn.save.plugins.fix
 
 import org.cqfn.save.core.config.TestConfig
 import org.cqfn.save.core.files.createFile
+import org.cqfn.save.core.files.createRelativePathToTheRoot
 import org.cqfn.save.core.files.myDeleteRecursively
 import org.cqfn.save.core.files.readLines
 import org.cqfn.save.core.logging.describe
@@ -27,6 +28,7 @@ import io.github.petertrr.diffutils.patch.Patch
 import io.github.petertrr.diffutils.text.DiffRowGenerator
 import okio.FileSystem
 import okio.Path
+import okio.Path.Companion.toPath
 
 import kotlin.random.Random
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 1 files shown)


save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -243,8 +245,8 @@ class FixPlugin(
         @Serializable(with = PathSerializer::class) val expected: Path
     ) : TestFiles {
         override fun withRelativePaths(root: Path) = copy(
-            test = test.relativeTo(root),
-            expected = expected.relativeTo(root),
+            test = test.createRelativePathToTheRoot(root).toPath(),
```
kgevorkyan, line 248: In https://github.com/square/okio/commit/a235589cac329973c4799ce9c6474e0b2ddb81e4#diff-eaa42bf2d76129a0f76061b7b82e8a904546f522cf2947b23207d0c010b7f1d6R260

they disable normalization in `commonRelativeTo` method (and this can't be configured), so we've got an extra `../` in our reporters
the calling of `normalized()` after `relativeTo` didn't change anything, I also didn't find any other canonization methods, so I restore our realization

petertrr: In [release notes](https://github.com/diktat-static-analysis/save-cloud/pull/501/commits/b71b58d614d2733e0be77629fe2eafc160f1e127) they state that `Path.normalize()` or `FileSystem.canonicalize()` can be used, have you tried these?

save-plugins/fix-plugin/src/commonMain/kotlin/org/cqfn/save/plugins/fix/FixPlugin.kt
```diff
@@ -243,8 +245,8 @@ class FixPlugin(
...
+            expected = expected.createRelativePathToTheRoot(root).toPath(),
```
petertrr, line 249: Can we cover this with any tests?

kgevorkyan: I tried the `normalize`, but it changes nothing, about `canonicalize` - it will remove extra `..`, but it will also append relative path from the root like
```bash
Test1Test.java -> C:\dir1\dir2\Test1Test.java
```

`createRelativePathToTheRoot` actually covered by tests, but in future, when `.relativeTo` from okio will work suitable for us, may be we can not just change `createRelativePathToTheRoot` to the `relativeTo`, but put
`relativeTo` into `createRelativePathToTheRoot`, since the latest is the wrapper, which is covered by tests

petertrr (approved):

PR #342, opened 2021-12-16
Added actual info in README.md

Cheshiriks:
What's done:
* Changes README


save-plugins/warn-plugin/README.md
```diff
@@ -27,7 +27,11 @@ Warning messages are very flexible and can be described in very different ways:
 ```
 // ;warn:3:1: Warning with an explicit set of a line number and column number
 ```
-
+```
+/* ;warn:1: Multiline warning.
+ * Requires you to specify a value of `expectedWarningsEndPattern` (and `expectedWarningsMiddlePattern` if there is)
+ */
```
petertrr, line 33:
```suggestion
/* ;warn:1: Multiline warning. This line should match `expectedWariningsPattern`
 * To match other lines, you need to specify a value of `expectedWarningsEndPattern` (and `expectedWarningsMiddlePattern` if there is any).
  * This line should match `expectedWarningsMiddlePattern`
  * And the next line should match `expectedWarningsEndPattern`
 */
```

Cheshiriks: Thank you) Fixed

PR #351, opened 2022-01-11
ci: Use jacoco to calculate code coverage from JVM tests

petertrr:
* Calculate code coverage using jacoco for JVM tests and common code
* Enable codecov upload in GH actions
* Disable `AVOID_NULL_CHECKS` in diktat

Some generated code still gets included into coverage reports, but only certain classes can be easily excluded, and generated methods can't.


diktat-analysis.yml
```diff
@@ -2,9 +2,11 @@
   enabled: true
   configuration:
     domainName: org.cqfn.save
-    kotlinVersion: 1.5
+    kotlinVersion: 1.6
     srcDirectories: "main,nativeMain,commonNonJsMain"
     testDirs: "test,nativeTest,commonTest,jvmTest,commonNonJsTest"
+- name: AVOID_NULL_CHECKS
+  enabled: false
```
kgevorkyan, line 9: May be some link to https://github.com/analysis-dev/diktat/issues/1167?

petertrr: It's not specifically because of issues, I just think that generally this rule overcomplicates code in SAVE, where we have a lot of branching. So simple comparisons with null look more clear. I think, we discussed this some time ago

kgevorkyan: ok

akuleshov7 (approved):

PR #354, opened 2022-01-17
Cmd execution mechanism refactoring

akuleshov7:


save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessBuilder.kt
```diff
@@ -105,7 +102,7 @@ class ProcessBuilder(private val useInternalRedirections: Boolean, private val f
 
         val cmd = modifyCmd(command, directory, processBuilderInternal)
 
-        logDebug("Executing: $cmd with timeout $timeOutMillis")
+        logInfo("Executing: $cmd with timeout $timeOutMillis")
```
akuleshov7, line 105: temp

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/CmdExecutorBase.kt
```diff
@@ -0,0 +1,103 @@
...
+    fun constructExecCmd(tmpDirName: String): String {
+        val execFlagsAdjusted = resolvePlaceholdersFrom(execFlags(), extractExtraFlags(), constructFileNamesForExecCmd(tmpDirName))
+        constructedExecCmd = "$execCmd $execFlagsAdjusted"
+        return constructedExecCmd
+    }
+
+    private fun constructFileNamesForExecCmd(tmpDirName: String): String {
+        // joining test files to string with a batchSeparator if the tested tool supports processing of file batches
+        // NOTE: SAVE will pass relative paths of Tests (calculated from testRootConfig dir) into the executed tool
+        val fileNamesForExecCmd = when (wildCardInDirectoryMode()) {
```
kgevorkyan, line 65: Just for interest, why do you prefer `when` here instead of `.let` construction?
We can have only two conditions: `null` and `not null`. And `let` copes with this task well, also here  you have double invocation of `wildCardInDirectoryMode()`, previously we used `it`

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/CmdExecutorBase.kt
```diff
@@ -0,0 +1,103 @@
...
+     */
+    fun executeCommandAndGetTestResults(redirectTo: Path?): ExecutionResult {
+        val time = timeOutMillis.times(copyPaths.size)
+        val execResult = pb.exec(constructedExecCmd, testConfig.getRootConfig().directory.toString(), redirectTo, time)
+
+        return ExecutionResult(execResult.code, execResult.getStdout(), execResult.stderr)
+    }
+
+    /**
+     * @param tmpDirName - the name of test file in the temporary directory
```
kgevorkyan, line 53: May be you can also add here few words about general idea and final view of `constructed cmd`? How it will be looks like and what we are going to do (how to modify) initial execCmd?

For now we put in `resolvePlaceholdersFrom` another 3 functions, which do some changes, which is quite loaded for understanding

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -123,125 +97,101 @@ class WarnPlugin(
...
-                    }
-                    "$testRootPath$it"
-                } ?: copyPaths.joinToString(separator = warnPluginConfig.batchSeparator!!)
+        val execCmd = cmdExecutor.constructExecCmd(tmpDirName)
 
-        logDebug("Constructed file name for execution for warn plugin: $fileNamesForExecCmd")
-        val execFlagsAdjusted = resolvePlaceholdersFrom(warnPluginConfig.execFlags, extraFlags, fileNamesForExecCmd)
-        val execCmd = "${generalConfig.execCmd} $execFlagsAdjusted"
-        val time = generalConfig.timeOutMillis!!.times(copyPaths.size)
+        try {
```
kgevorkyan, line 127: Why do we need `try` on this whole block, while catching only `pb` exceptions? Could it be done only for
`cmdExecutor.executeCommandAndGetTestResults` like before?

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/CmdExecutorBase.kt
```diff
@@ -0,0 +1,103 @@
...
+
+        return extraFlagsList.singleOrNull() ?: ExtraFlags("", "")
+    }
+
+    /**
+     * @param redirectTo a file where process output and errors should be redirected.
+     * If null, output will be returned as [ExecutionResult.stdout] and [ExecutionResult.stderr].
+     * @return the result after the execution of [constructedExecCmd]
+     */
+    fun executeCommandAndGetTestResults(redirectTo: Path?): ExecutionResult {
```
kgevorkyan, line 45: We have two different classes in SAVE: `TestResult` and `ExecutionResult`. may be in aim to avoid misunderstanding we can rename this method

```suggestion
    fun executeCommandAndGetExecutionResults(redirectTo: Path?): ExecutionResult {
```

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
akuleshov7: unfortunately not, yeah, I tried to refactor this in different ways. 
But we need to set `execCmd` in case of exception in catch block, so it cannot be handled inside :(

This is the best variant unfortunately

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/CmdExecutorBase.kt
akuleshov7: fixed

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/CmdExecutorBase.kt
akuleshov7: fixed

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/CmdExecutorBase.kt
akuleshov7: fixed

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
akuleshov7: fixed

kgevorkyan (approved):

PR #359, opened 2022-02-11
Support reading of actual warning in SAVE from the file

Cheshiriks:
What's done:
* Added reading of actual warning from the file

### DIFF
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
@@ -147,12 +147,10 @@ class WarnPlugin(
             return failTestResult(originalPaths, ex, execCmd)
         }
 
-        val actualWarningsMap = if (warnPluginConfig.actualWarningsFileName != null) {
+        val actualWarningsMap = warnPluginConfig.actualWarningsFileName?.let {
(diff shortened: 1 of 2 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -146,7 +147,12 @@ class WarnPlugin(
             return failTestResult(originalPaths, ex, execCmd)
         }
 
-        val actualWarningsMap = collectActualWarningsWithLineNumbers(result, warnPluginConfig)
+        val actualWarningsMap = if (warnPluginConfig.actualWarningsFileName != null) {
```
akuleshov7, line 150: may be `let` here (as suggested in diktat)?

akuleshov7 (approved):

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
Cheshiriks: Yes, fixed

Cheshiriks:
> What about test?

It is there, but I did not add it because of the full path to the file

kgevorkyan:
If `okio` still not support `absolute`, let's at least create draft PR with your test, after merging of this one

kgevorkyan (approved):

Cheshiriks:
> If `okio` still not support `absolute`, let's at least create draft PR with your test, after merging of this one

Ok

PR #380, opened 2022-04-11
Use sarif4k from analysis-dev GH packages

petertrr:
Instead of `includeBuild`, use artifacts published from https://github.com/analysis-dev/sarif4k/tree/feature/publishing to Github Package registry.
With this change developers will have to add additional properties to work with save-cli source code:
```properties
gprUser=<GH username>
gprKey=<GH personal access token>
```
Personal Access Token should be generated via https://github.com/settings/tokens/new with the scope at least containing `read:packages`.
This, however, is a temporary workaround, which should be more fast and stable than `includeBuild`

### DIFF
--- a/save-plugins/warn-plugin/build.gradle.kts
+++ b/save-plugins/warn-plugin/build.gradle.kts
@@ -8,7 +8,7 @@ kotlin {
             dependencies {
                 implementation(projects.saveCommon)
                 implementation(libs.kotlinx.serialization.core)
-                implementation("io.github.detekt.sarif4k:sarif4k")
+                implementation("io.github.detekt.sarif4k:sarif4k:0.1.0-SNAPSHOT")
             }
         }
     }
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -7,11 +7,22 @@ include("save-plugins:fix-plugin")
 include("save-plugins:warn-plugin")
 include("save-reporters")
 include("save-common-test")
-includeBuild("sarif4k")
 
 dependencyResolutionManagement {
     repositories {
         mavenCentral()
+        maven {
+            url = uri("https://maven.pkg.github.com/analysis-dev/sarif4k")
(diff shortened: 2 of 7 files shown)


akuleshov7:
Actions hanged...

save-plugins/warn-plugin/build.gradle.kts
```diff
@@ -8,7 +8,7 @@ kotlin {
             dependencies {
                 implementation(projects.saveCommon)
                 implementation(libs.kotlinx.serialization.core)
-                implementation("io.github.detekt.sarif4k:sarif4k")
+                implementation("io.github.detekt.sarif4k:sarif4k:0.1.0-SNAPSHOT")
```
akuleshov7, line 11: snapshot?

settings.gradle.kts
```diff
@@ -7,11 +7,22 @@ include("save-plugins:fix-plugin")
...
 include("save-reporters")
 include("save-common-test")
-includeBuild("sarif4k")
 
 dependencyResolutionManagement {
     repositories {
         mavenCentral()
+        maven {
+            url = uri("https://maven.pkg.github.com/analysis-dev/sarif4k")
+            val gprUser: String? by settings
```
akuleshov7, line 16: please notify everyone in our team about this change, that they need to set tokens, ok?

akuleshov7 (approved):

petertrr:
> Actions hanged...

Yeah, that's because I've added `[skip ci]` in the last commit, because there were only updates in documentation

save-plugins/warn-plugin/build.gradle.kts
petertrr: Currently released version of `sarif4k` is `0.0.1` and the next development version in our fork is `0.1.0-SNAPSHOT`. I thought we shouldn't create release versions in GH packages (yet).

PR #381, opened 2022-04-14
Update diktat to 1.1.0

petertrr:
* Enable githubActions mode for CI
* Fix diktat findings
* Remove diktatCheckAll and diktatFixAll


save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
```diff
@@ -116,7 +116,7 @@ abstract class Plugin(
                     }
                 }
             }.toList()
-            val notFoundTests = testFiles.filter { it !in foundTests.map { foundTest -> foundTest.test.toString() } }
+            val notFoundTests = testFiles.filter { testFile -> testFile !in foundTests.map { foundTest -> foundTest.test.toString() } }
```
kgevorkyan, line 119: maybe it will be good to have logic on number of maximum inline lambdas in diktat

kgevorkyan (approved):

save-common/src/commonMain/kotlin/org/cqfn/save/core/plugin/Plugin.kt
petertrr: Yes, I agree. This logic could be added into `NewlinesRule` as a new inspection. Feel free to open an issue in diktat.

PR #383, opened 2022-04-20
Update okio to 3.1.0, enable HMPP

### LINKED ISSUE #292: Enable HMPP structure after okio supports it
According to [docs](https://kotlinlang.org/docs/mpp-share-on-platforms.html#share-code-on-similar-platforms), several properties need to be enabled to use platform-dependent functions from shared source sets (like `nativeMain`) so that IDE correctly resolves them. However, okio doesn't support them (https://github.com/square/okio/pull/980 is required), so we can't enable it either.

petertrr:
* Enable HMPP
* Fix source set dependencies (`*Test` automatically depends on `*Main`, [explanation](https://youtrack.jetbrains.com/issue/KT-47144/[Multiplatform]-Setting-*Test.dependsOn(*Main)-dependency-betwee#focus=Comments-27-5097081.0-0))
* Make `fs` `expect` in save-common-test, because it doesn't work otherwise (there is no `expect val SYSTEM` in okio; `FileSystem.SYSTEM` just appears in some of target source sets. See https://github.com/square/okio/discussions/1115 for more details.)
* Substitute `getcwd` with okio-based implementation, because apparently `getcwd` is not present in common posix for 1.6.20. I can't see any changes at https://github.com/JetBrains/kotlin/blob/master/kotlin-native/platformLibs/src/platform/mingw/posix.def, but according to https://www.gnu.org/software/gnulib/manual/html_node/getcwd.html there are some differences listed between Linux and Mingw. Not sure, why it worked before though.
* Fix condition in diktat workflow

Overall the code everything seems to be resolved better by IDEA, there is much less red.
  Closes #292

### DIFF
--- a/save-common/src/nativeMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
+++ b/save-common/src/nativeMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
@@ -36,5 +36,4 @@ actual fun FileSystem.myDeleteRecursively(path: Path) {
     }, 64, FTW_DEPTH)
 }
 
-@Suppress("GENERIC_VARIABLE_WRONG_DECLARATION")
 actual fun getWorkingDirectory(): Path = fs.canonicalize(".".toPath())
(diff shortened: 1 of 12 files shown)


save-common/src/nativeMain/kotlin/org/cqfn/save/core/files/FileUtils.kt
```diff
@@ -41,11 +37,4 @@ actual fun FileSystem.myDeleteRecursively(path: Path) {
 }
 
 @Suppress("GENERIC_VARIABLE_WRONG_DECLARATION")
```
kgevorkyan, line 39: This probably can be removed too

kgevorkyan (approved):

PR #390, opened 2022-05-24
Parsing exceptions during the parsing of invalid SARIF result

### LINKED ISSUE #388: Parsing exceptions during the parsing of invalid SARIF result
If the tool provides us invalid SARIF result and we fail on parsing of it - we know it definitely.

We need to handle that situation and NOT crush save-cli with exception. Instead - we need to catch an error and mark all tests in the batch **FAILED** with the reason in the debug info: `WE FAILED TO PARSE SARIF!!!111111 WOW, YOUR TOOL IS INVALID, CHECK THE GENERATION OF SARIF REPORT`.

Cheshiriks:
* Added exception for parsing sarif
Closes #388

### DIFF
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessParseSarifException.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessParseSarifException.kt
@@ -1,0 +1,6 @@
+package org.cqfn.save.core.utils
+
+/**
+ * An [Exception] that can be thrown in case of failed to parse sarif
+ */
+open class ProcessParseSarifException(message: String) : Exception(message)
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/SarifParsingException.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/SarifParsingException.kt
@@ -3,4 +3,4 @@ package org.cqfn.save.core.utils
 /**
  * An [Exception] that can be thrown in case of failed to parse sarif
  */
-class SarifParsingException(message: String) : Exception(message)
+class SarifParsingException(message: String, cause: Throwable?) : Exception(message, cause)
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/SarifParsingException.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/SarifParsingException.kt
@@ -3,4 +3,4 @@ package org.cqfn.save.core.utils
 /**
  * An [Exception] that can be thrown in case of failed to parse sarif
  */
-open class ProcessParseSarifException(message: String) : Exception(message)
+class SarifParsingException(message: String) : Exception(message)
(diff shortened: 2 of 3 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
```diff
@@ -206,14 +219,19 @@ class WarnPlugin(
...
         warnPluginConfig: WarnPluginConfig,
         originalPaths: List<Path>,
         copyPaths: List<Path>,
         workingDirectory: Path,
     ): WarningMap = if (warnPluginConfig.expectedWarningsFormat == ExpectedWarningsFormat.SARIF) {
-        val warningsFromSarif = collectWarningsFromSarif(warnPluginConfig, originalPaths, fs, workingDirectory)
+        val warningsFromSarif = try {
+            collectWarningsFromSarif(warnPluginConfig, originalPaths, fs, workingDirectory)
+        } catch (e: Exception) {
+            throw ProcessParseSarifException("We failed to parse sarif. Your tool is invalid, check the generation of sarif report")
```
petertrr, line 233: Why are you ignoring the original exception? It contains very valuable information for the user. Was it deserialization exception, i.e. SARIF is malformed? Or was it an IO exception, e.g. file couldn't be opened for some reason? Please include it as a cause of this exception and check whether we discard it sometime later. Users should be able to see this information.
Also - wording `you tool is invalid` is too harsh :)

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessParseSarifException.kt
```diff
@@ -0,0 +1,6 @@
+package org.cqfn.save.core.utils
+
+/**
+ * An [Exception] that can be thrown in case of failed to parse sarif
+ */
+open class ProcessParseSarifException(message: String) : Exception(message)
```
petertrr, line 6:
```suggestion
open class SarifParsingException(message: String) : Exception(message)
```
Also, why `open class`?

save-plugins/warn-plugin/src/commonMain/kotlin/org/cqfn/save/plugin/warn/WarnPlugin.kt
Cheshiriks: Ok. Fixed:)

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/ProcessParseSarifException.kt
Cheshiriks: Fixed

save-common/src/commonMain/kotlin/org/cqfn/save/core/utils/SarifParsingException.kt
```diff
@@ -0,0 +1,6 @@
+package org.cqfn.save.core.utils
+
+/**
+ * An [Exception] that can be thrown in case of failed to parse sarif
+ */
+class SarifParsingException(message: String) : Exception(message)
```
petertrr, line 6: Please also add `cause` as a parameter (`Exception` has a constructor that accepts cause). We may want to use it in the future for better reporting, so it'll be useful to have it in there already

Cheshiriks: Added

PR #396, opened 2022-06-01
save-cli: Precision and recall

nulls:
### What's done:
- renamed match to matched
- renamed missing to unmatched

it's part of https://github.com/saveourtool/save-cloud/issues/514


akuleshov7 (approved):
lgtm, for sure

PR #404, opened 2022-06-22
Fix inheritance of loading configs

### LINKED ISSUE #403: SAVE-cli incorrectly uses inheritance
When we run `./save warn-dir/chapter2` from `example/kotlin-diktat` directory, diktat fails with the following error:
`no exec_cmd in config`. It looks like the inheritance of configs is done incorrectly

nulls:
### What's done:
- detect *.properties
- load parent config

It closes #403


save-cli/src/nativeMain/kotlin/com/saveourtool/save/cli/logging/Logger.kt
```diff
@@ -16,7 +16,7 @@ import kotlin.system.exitProcess
  * @return nothing, program terminates in this method
  */
 @Deprecated("never use this method in save-core as it can lead to a break of save-cloud application")
-fun logErrorAndExit(exitCode: ExitCodes, message: String): Nothing {
+fun <R> logErrorAndExit(exitCode: ExitCodes, message: String): R {
```
nulls, line 19: added some type to be able to use it as return statement

save-cli/src/nativeMain/kotlin/com/saveourtool/save/cli/SaveConfigReader.kt
```diff
@@ -130,6 +124,55 @@ fun readPropertiesFile(propertiesFileName: String?): SaveProperties {
     return Properties.decodeFromStringMap(serializer(), properties)
 }
 
+private fun detectPropertiesFile(testRootDirectories: List<String>?): Path? {
```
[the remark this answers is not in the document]
nulls: split this property to two fields in `SaveProperties` class

buildSrc/src/main/kotlin/com/saveourtool/save/generation/Generation.kt
```diff
@@ -87,40 +117,53 @@ fun FunSpec.Builder.generateOptions(jsonObject: Map<String, Option>): FunSpec.Bu
...
+ * @param arguments map of cli argument names to [Argument] objects
  * @return a corresponding [FunSpec.Builder]
  */
-fun FunSpec.Builder.assignMembersToOptions(jsonObject: Map<String, Option>): FunSpec.Builder {
-    jsonObject.forEach {
-        val assign = "this.${it.key} = ${it.key}"
-        this.addStatement(assign)
+fun FunSpec.Builder.assignMembers(options: Map<String, Option>, arguments: Map<String, Argument>): FunSpec.Builder {
+    options.forEach { (key, _) ->
+        this.addStatement("$key = $key ?: defaultProperties.$key,")
```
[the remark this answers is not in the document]
nulls: done

save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/utils/CliUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+         * Parse properties file ([rootDir]/[projectName].properties) to [T] structure
+         *
+         * @param fs implementation of [FileSystem]
+         * @param rootDir
+         * @param projectName
+         * @return parsed [T] structure
+         * @throws IOException when [rootDir] not found
+         */
+        @OptIn(ExperimentalSerializationApi::class)
+        inline fun <reified T> parsePropertiesFile(
```
petertrr, line 28: Why not make these utils top-level methods? I actually think we had a rule in Diktat that marks classes only with methods as an anti-pattern

save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/TestConfig.kt
```diff
@@ -122,12 +122,15 @@ data class TestConfig(
      * @return an update this [TestConfig]
      */
     fun processInPlace(createPluginConfigList: (TestConfig) -> List<PluginConfig>): TestConfig {
+        // need to process parent first
+        this.parentConfig?.processInPlace(createPluginConfigList)
```
petertrr, line 126: Why is this needed here? Shouldn't we start iterating from parent here: https://github.com/saveourtool/save-cli/blob/93a615e0f0c5b48e386ff86e8485e59e35d4bcb1/save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/Save.kt#L71-L74
And have parent config already processed by this time?

nulls: it's required when testRootDir is sub folder, for example:
1. `./save.toml` contains information how to run
2. run tests `save-cli ./subfolder/ ...`.
3. save-cli processes `./subfolder/save.toml`, but it doesn't contain information how to run the test (`no exec_cmd in config`)

save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/utils/CliUtils.kt
nulls: It was done to be able to use it via class name: `CliUtils.parsePropertiesFile()` instead of just `parsePropertiesFile()`. For me, it increases code readable (you can see class for called method from method's call without checking imports).

But if there is a rule which prohibits public methods in object companion -- will change it to top level methods

save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/TestConfig.kt
nulls: will add a test with example which was provided by @akuleshov7 for original issue

petertrr: Yes, but I thought that when we call `configFromFile() -> discoverConfigWithParents -> getTestConfigFromTomlFile` we are processing all parent directories of the current file, so it seems that the original intention was to process them there. Probably, there is a bug in there somewhere.

petertrr (approved):

save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/TestConfig.kt
nulls: discussed offline -- raised https://github.com/saveourtool/save-cli/issues/415 for it
probably need to fix argument descriptions

save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/utils/CliUtils.kt
nulls: it's about extension for class, will refactor to extension for `FileSystem`

Issue #409, opened 2022-06-24
Improved logic of directory mode in warn plugin


petertrr:
As we discussed: 
1) [warn] plugin -> is simply processing separate files
2) [warn] plugin + DIRECTORY flag -> run ONLY on this directory. Plugin uses from one single file with all warnings (no inline warnings). It will make the logic less complex. In this mode (with directory flag) we will ignore other nested save.toml files and detect tests from the file with expected warnings.

We need to create a logic on save-cloud for it also...

_Originally posted by @akuleshov7 in https://github.com/saveourtool/save-cli/issues/402#issuecomment-1165571071_

petertrr:
Another related comment from that thread: https://github.com/saveourtool/save-cli/issues/402#issuecomment-1164620902.

And one more suggestion - we discussed that we should extract test discovering logic into an interface with several implementations: match by suffix (default behavior of warn plugin), match a pair of files by suffix (fix plugin) and match all files in a directory (new logic). However, then we decided to remove discovering in directory mode and rely only on a file with a list of expected tests. Nevertheless, extracting logic is almost always a good idea :)

PR #413, opened 2022-06-27
set test location relative path relative to original test path

icemachined:
to be in sync with save-cloud logic

### DIFF
--- a/save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/Save.kt
+++ b/save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/Save.kt
@@ -48,6 +48,12 @@ class Save(
     internal val reporter = getReporter(saveProperties)
 
     /**
+     * test root path which is set by user and can be different from config root path
+     * it can contain only subset of the whole test set
+     * */
+    private val testRootPath = saveProperties.testRootDir.toPath()
+
+    /**
      * Main entrypoint for SAVE framework. Discovers plugins and calls their execution.
      *
      * @return Reporter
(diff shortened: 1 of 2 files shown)


save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/Save.kt
```diff
@@ -56,7 +57,6 @@ class Save(
     fun performAnalysis(): Reporter {
         logInfo("Welcome to SAVE version $SAVE_VERSION")
         // FixMe: now we work only with the save.toml config and it's hierarchy, but we should work properly here with directories as well
```
petertrr, line 59: I think this comment is already outdated and can be removed. Logic with directories is partially implemented and will be improved under #409

petertrr (approved):
So with this change we'll have full paths in save-cloud execution history? I.e. `examples/kotlin-diktat/warn/chapter1` where we now have just `warn/chapter1`?

nulls (approved):

nulls (changes requested):
please rebase with latest changes and check a new version of `SaveProperties`

save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/Save.kt
```diff
@@ -147,12 +151,10 @@ class Save(
         logDebug("=> Executing plugin: ${plugin::class.simpleName} for [${plugin.testConfig.location}]")
         reporter.onPluginExecutionStart(plugin)
         try {
-            val testRepositoryRootPath = plugin.testConfig.getRootConfig().location
-
             plugin.execute()
                 .onEach { event ->
                     // calculate relative paths, because reporters don't need paths higher than root dir
-                    val resourcesRelative = event.resources.withRelativePaths(testRepositoryRootPath)
```
nulls, line 155: previously we took it from plugin -- do you sure that it has the same value as from `saveProperties`?

icemachined: here we change logic to be compliant with the same logic from save-cloud, in save cloud it expect tests in folders which are  relative to testRootPath, when this path is not the same as test repository root path when we cant download debug info in gui

nulls (approved):

Issue #312, opened 2021-11-17
SAVE shouldn't mutate sources it operates on


petertrr:
Now after FixPlugin execution sources have different line separators (`\r\n` becomes `\n`)

akuleshov7:
Very similar to https://github.com/saveourtool/save-cli/issues/280 
> When expected file is re-written, line separators are changed on Windows.

PR #417, opened 2022-06-28
Refactoring of gradle scripts to add 'jvm' target to cli

akuleshov7:


save-cli/build.gradle.kts
```diff
@@ -71,6 +86,13 @@ kotlin {
         dependsOn(linkReleaseExecutableTaskProvider)
     }
 
+    // disable building of some binaries to speed up build
+    // possible values: `all` - build all binaries, `debug` - build only debug binaries
+    val enabledExecutables = if (hasProperty("enabledExecutables")) property("enabledExecutables") as String else null
+    if (enabledExecutables != null && enabledExecutables != "all" || enabledExecutables == "debug") {
```
akuleshov7, line 92: Now I get: `Condition 'enabledExecutables == "debug"' is always false when reached`

I do not know how it was working previously...

petertrr: Do you set `-PenabledExecutables` to something other?

akuleshov7: > Do you set `-PenabledExecutables` to something other?

@petertrr Idea raises this warning without setting any parameters 
image

petertrr: Yep, because it's obviously a sub-case of the first part: `A != X || A == Y` 🤔 So, we are currently only distinguishing `-PenabledExecutables=all`

examples/kotlin-diktat/fix_and_warn/smoke/src/main/kotlin/com/saveourtool/save/chapter3/Example1Expected.kt
```diff
@@ -1,10 +1,6 @@
-// ;warn:1:1 [FILE_NAME_MATCH_CLASS] file name is incorrect - it should match with the class described in it if there is the only one class declared: Example1Expected.kt vs Example{{.*}}
```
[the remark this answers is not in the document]
petertrr: Related to #312?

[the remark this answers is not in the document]
akuleshov7: No, I think it happened because I have interrupted build with ctrl + C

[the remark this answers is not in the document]
petertrr: Yes, but I don't think we touch files with Gradle. It must've been done either by our tests or by save itself

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/JacocoConfiguration.kt
```diff
@@ -49,6 +43,19 @@ fun Project.configureJacoco() {
             html.required.set(true)
         }
     }
+
+    // `application` plugin creates jacocoTestReport task in plugin section (this is definitely incorrect behavior)
+    // AFTER that in "com.saveourtool.save.buildutils.kotlin-library" we try to register this task once again and fail
+    // so the order of plugins in `apply` is critically important
+    val jacocoTestReportTask = if (project.name == "save-cli") {
```
akuleshov7, line 50: @petertrr any ideas why this can happen? I am completely lost :(
```

* What went wrong:
An exception occurred applying plugin request [id: 'com.saveourtool.save.buildutils.kotlin-library']
> Failed to apply plugin 'com.saveourtool.save.buildutils.kotlin-library'.
   > Extension of type 'JacocoReport' does not exist. Currently registered extension types: [ExtraPropertiesExtension]

```

Also funny thing - that plugins order is important in our logic, because we cannot manipulate with the creation of `jacocoTestReport` in `application` plugin

petertrr: > `application` plugin creates jacocoTestReport task in plugin section (this is definitely incorrect behavior)

Why? Gradle calls plugins `apply` method and the plugin performs all its necessary setup, which includes registering of tasks.

> `> Extension of type 'JacocoReport' does not exist. Currently registered extension types: [ExtraPropertiesExtension]`

Looks like jacoco plugin hasn't been applied yet. Do you get this error in branch with `register` for `jacocoTestReport` task? Then may be we pass `configure` incorrectly and it's not lazily evaluated?

akuleshov7: >>> Do you get this error in branch with register for jacocoTestReport task?
Nope, I get it on `save-cli` (on the branch where we use jacocoTestReport task created in `application` plugin)

petertrr: `jacoco` plugin reacts to the presence of `java` plugin (which is applied by `application` plugin) as described [here](https://docs.gradle.org/current/userguide/jacoco_plugin.html#sec:jacoco_tasks). I think `getByName` is an incorrect choice [after all](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:how_do_i_reference_a_task), because it'll get task immediately and call `configure` on the task instead of a `TaskProvider`. I think more appropriate would be to use `tasks.named` and then call `configure` on the returned provider.

akuleshov7: [How do I reference a task without creating/configuring it?](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:how_do_i_reference_a_task)

Instead of referencing a task object, you can work with a registered task via a [TaskProvider](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskProvider.html) object. A [TaskProvider](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskProvider.html) can be obtained in several ways including when calling [TaskContainer.register(java.lang.String)](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskContainer.html#register-java.lang.String-) or using the [TaskCollection.named(java.lang.String)](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskCollection.html#named-java.lang.String-) method.

Yes, this is what they suggest

akuleshov7: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview

akuleshov7: If you are configuring a task by name, you will need to use the configuration avoidance equivalent. See the [table below](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview) to identify the alternative.

akuleshov7: In TaskProvider they have [configure](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskProvider.html#configure-org.gradle.api.Action-)​([Action](https://docs.gradle.org/current/javadoc/org/gradle/api/Action.html)<? super [T](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskProvider.html)> action)

petertrr: But `register` returns `TaskProvider` too, so using `named` should only make two branches more similar

akuleshov7: yeah, but that API without examples was very confusing (or may be I am too stupid for gradle)

examples/kotlin-diktat/fix_and_warn/smoke/src/main/kotlin/com/saveourtool/save/chapter3/Example1Expected.kt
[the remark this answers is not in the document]
petertrr: Shouldn't this change be restored BTW?

PR #422, opened 2022-07-07
Set Gradle property to enable Kotlin XML test report

petertrr:
In #421 we introduced a separate step to re-run JVM tests, because by default when tests for all targets are executed, Kotlin Gradle Plugin generates only a single HTML report. This behavior can be turned off, however.

There is a more-or-less detailed discussion at https://youtrack.jetbrains.com/issue/KT-32608/Create-JUnit-XML-result-file-in-multiplatform-gradle-build

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -36,29 +36,15 @@ jobs:
       - uses: gradle/gradle-build-action@v2
         with:
           gradle-version: wrapper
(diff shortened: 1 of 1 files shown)


petertrr:
@0x6675636b796f75676974687562 do we really need `dorny/test-reporter` action? Looks like it generates the same information as an entry in `Build and test` workflow with only difference that it can be opened from checks tab.

PR #437, opened 2022-08-23
Introduced `not applicable` counter for FixPlugin

nulls:
### What's done:
* added a new constant for CountWarnings

It closes https://github.com/saveourtool/save-cloud/issues/1039

### DIFF
--- a/save-common/src/commonMain/kotlin/com/saveourtool/save/core/result/DebugInfo.kt
+++ b/save-common/src/commonMain/kotlin/com/saveourtool/save/core/result/DebugInfo.kt
@@ -37,7 +37,7 @@ data class CountWarnings(
     val unexpected: Int,
 ) {
     companion object {
-        const val NOT_APPLICABLE_COUNTER: Int = -99
+        private const val NOT_APPLICABLE_COUNTER: Int = -99
 
         /**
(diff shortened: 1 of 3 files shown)


save-common/src/commonMain/kotlin/com/saveourtool/save/core/result/DebugInfo.kt
```diff
@@ -35,4 +35,18 @@ data class CountWarnings(
     val matched: Int,
     val expected: Int,
     val unexpected: Int,
-)
+) {
+    companion object {
+        const val NOT_APPLICABLE_COUNTER: Int = -99
```
petertrr, line 40: Maybe also add method to check if DebugInfo or a particular counter is `not applicable`? To fully encapsulate this logic here and only call this method from sae-coud

petertrr (approved):

save-common/src/commonMain/kotlin/com/saveourtool/save/core/result/DebugInfo.kt
nulls: good point, will do

nulls: done

PR #438, opened 2022-08-25
Added `save-overrides.toml`

### LINKED ISSUE #435: Introduce `save-overrides.toml`
This file should contain following options:

- execCmd (or initially overrideExecCmd)
- execFlags (before https://github.com/saveourtool/save-cli/issues/434, later can be skipped)
- ~batchSeparator~
- ~batchSize~

As bonus - should be done separately later: 

`This file should support groups `[fix]` and `[warn]` to customize execution run for different type of plugins`

It allows to store such files in private repositories for test and provide them as external file for public tests.

_Originally posted by @nulls in https://github.com/saveourtool/save-cli/issues/430#issuecomment-1217718558_

nulls:
Supported a new file `save-overrides.toml`. It has the same structure as `save.toml`, but we expect it only in root for current run.

Also moved `EvaluatedToolConfig` to `TestConfig`

It closes #435


save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/TestConfig.kt
```diff
@@ -122,15 +130,17 @@ data class TestConfig(
      * @return an update this [TestConfig]
      */
     fun processInPlace(createPluginConfigList: (TestConfig) -> List<PluginConfig>): TestConfig {
-        // need to process parent first
-        this.parentConfig?.processInPlace(createPluginConfigList)
```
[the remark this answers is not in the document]
nulls: yep, moved to `processWithParentsInPlace` and it's called for parent of root `TestConfig`

save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/EvaluatedToolConfig.kt
```diff
@@ -1,16 +1,12 @@
...
+ * Configuration for an evaluated tool, that is read from cli
  *
- * @property execCmd
- * @property execFlags
  * @property batchSize it controls how many files execCmd will process at a time
  * @property batchSeparator A separator to join test files to string if the tested tool supports processing of file batches (`batch-size` > 1)
  */
 data class EvaluatedToolConfig(
-    val execCmd: String?,
-    val execFlags: String?,
```
petertrr, line 13: What about those properties? They are removed because they can be set either from save.toml or from ~save-overrides.toml~ CLI flags? Isn't it overcomplication to configure the evaluated tool from three places?

nulls: `execCmd` and `execFlags` now can be set only in `save.toml` (in future it will be combined to a single field) -- personal tests and can be overridden by `save-overrides.toml` -- public\sharable tests

petertrr: Then what's the logic behind keeping batch size a a CLI flag? Maybe we should move everything to the previous logic with TOML files only?

nulls: actually, I partially agree: I personally think that we should extract all configs are related to tool from `save.toml`, but I understand the use case with private tests when execCmd is stored inside for ease usage and the case when execFlags changes from one test suite to another test suite.

@akuleshov7, what do you think? maybe return `batchSize` and `batchSeparator` back to `save.toml` but to general section?
Motivation: we don't remove `execCmd` and `execFlags` from `save.toml`, let's store `batchSize` and `batchSeparator` there

nulls: discussed -- agreed to move to `GeneralConfig`

akuleshov7: totally agree, I would even remove all non-critical arguments from cli

save-core/src/commonNonJsMain/kotlin/com/saveourtool/save/core/utils/TestConfigUtils.kt
```diff
@@ -20,6 +20,16 @@ fun TestConfig.processInPlace() = processInPlace {
     createPluginConfigListFromToml(it.location, fs)
 }
 
+/**
+ * Process current and all parents of [this] config, reading individual plugin configurations from TOML file.
+ *
+ * @return [this] config with all descendants being evaluated (`pluginConfigs` are filled with data and merged with parents)
+ */
+fun TestConfig.processWithParentsInPlace(): TestConfig {
+    parentConfig?.processInPlace()
```
kgevorkyan, line 29: > Process current and **all** parents

Shouldn't it like this be in aim to process all parents?

```suggestion
    parentConfig?.processWithParentsInPlace()
```

nulls: yes, good catch

nulls: fixed + added a test for it

petertrr (approved):

PR #443, opened 2022-08-30
Support escaping commas in extra flags

petertrr:


save-common/src/commonMain/kotlin/com/saveourtool/save/core/plugin/ExtraFlagsExtractor.kt
```diff
@@ -29,7 +29,8 @@ class ExtraFlagsExtractor(private val generalConfig: GeneralConfig,
      * @return [ExtraFlags] or null if no match occurred
      */
     internal fun extractExtraFlagsFrom(line: String) = line
-        .split(",", ", ")
+        .splitByNonEscaped(',')
+        .map { it.replace("\\,", ",") }
```
akuleshov7, line 33: case with escaping of a single slash won't work here

akuleshov7 (approved):

PR #446, opened 2022-09-02
Don't fail tests on Windows if `save.exe` is locked by an external process

0x6675636b796f75676974687562:


petertrr (approved):
Weird that it's file not found and not something about file access, but as long as it works

PR #447, opened 2022-09-02
Update the ignore patterns

0x6675636b796f75676974687562:


examples/kotlin-diktat/.gitignore
```diff
@@ -0,0 +1,3 @@
+/ktlint
```
petertrr, line 1: Maybe also add `diktat-*.jar`, or even simply `*.jar`?

petertrr: Oh, they are added to the root `.gitignore` already

petertrr (approved):

PR #464, opened 2022-11-29
Adapt FixPlugin for library for applying of SARIF fix patches

kgevorkyan:
### What's done:
* Adapt FixPlugin for library for applying of SARIF fix patches

FixAndWarnPlugin use FixPlugin for fixing test files and comparing with expected, which are released of warnings at the moment of FixPlugin start, so there is no need to make changes in `FixAndWarnPlugin`

### DIFF
--- a/save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/FixFormats.kt
+++ b/save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/FixFormats.kt
@@ -5,10 +5,10 @@
 package com.saveourtool.save.core.config
 
 /**
- * Possible formats of actual (i.e. produces by the tool) warnings
+ * Possible formats of actual set of fixes, provided by user
  */
 enum class ActualFixFormat {
-    PLAIN,
+    IN_PLACE,
     SARIF,
     ;
 }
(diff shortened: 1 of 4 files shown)


save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/FixFormats.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+@file:Suppress("FILE_NAME_MATCH_CLASS", "MatchingDeclarationName")
+
+package com.saveourtool.save.core.config
+
+/**
+ * Possible formats of actual set of fixes, provided by user
+ */
+enum class ActualFixFormat {
```
akuleshov7, line 12: is this flag used in fix and warn?

akuleshov7 (approved):

save-common/src/commonMain/kotlin/com/saveourtool/save/core/config/FixFormats.kt
kgevorkyan: `FixAndWarn` plugin actually have no own configuration, it use `fix` and `warn` configs as subsections, so there is no need of this flag in `FixAndWarn` directly. If it used in `fix` section, then it will be used in `FixAndWarn`

PR #468, opened 2022-12-22
fix `renovate` for `diktat-gradle-plugin`

nulls:


akuleshov7 (approved):

nulls:
`windows-latest` fails is because of gradle upgrade

nulls:
> `windows-latest` fails is because of gradle upgrade

rerun did help

PR #478, opened 2023-01-16
FixPlugin: Compare test files in course of building the TestResults by full paths

### LINKED ISSUE #473: FixPlugin: Compare test files in course of building the TestResults by full paths
When building test results at the end of work of `FixPlugin` we should compare test files not only by names, but by full paths, relative to the `testRoot`, since it might be several files with the same name, if `batchSize > 1`. Thus, we cant's just take the `first` one of them

https://github.com/saveourtool/save-cli/blob/a846ce9cd19b35f6984a6413896eac1dd4c134f9/save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt#L132

Cheshiriks:
* Added test for tests with the same name 

Closes #473

### DIFF
--- a/save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
+++ b/save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
@@ -230,10 +230,7 @@ class FixPlugin(
     ): List<TestResult> = testCopyToExpectedFilesMap.map { (testCopy, expected) ->
         val fixedLines = fs.readLines(testCopy)
         val expectedLines = fs.readLines(expected)
-
-        // FixMe: https://github.com/saveourtool/save-cli/issues/473
-        val test = testToExpectedFilesMap.first { (test, _) -> test.name == testCopy.name }.first
-
+        val test = testToExpectedFilesMap.first { (test, _) -> test.createRelativePathToTheRoot(testConfig.directory) in testCopy.toString() }.first
         TestResult(
             FixTestFiles(test, expected),
             checkStatus(expectedLines, fixedLines),
--- a/save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
+++ b/save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
@@ -5,6 +5,7 @@ import com.saveourtool.save.core.config.TestConfig
 import com.saveourtool.save.core.files.createFile
 import com.saveourtool.save.core.files.createRelativePathToTheRoot
 import com.saveourtool.save.core.files.myDeleteRecursively
+import com.saveourtool.save.core.files.parents
 import com.saveourtool.save.core.files.readLines
 import com.saveourtool.save.core.logging.describe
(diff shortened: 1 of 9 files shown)


save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
```diff
@@ -230,10 +230,7 @@ class FixPlugin(
     ): List<TestResult> = testCopyToExpectedFilesMap.map { (testCopy, expected) ->
         val fixedLines = fs.readLines(testCopy)
         val expectedLines = fs.readLines(expected)
-
-        // FixMe: https://github.com/saveourtool/save-cli/issues/473
-        val test = testToExpectedFilesMap.first { (test, _) -> test.name == testCopy.name }.first
-
+        val test = testToExpectedFilesMap.first { (test, _) -> test.createRelativePathToTheRoot(testConfig.directory) in testCopy.toString() }.first
```
kgevorkyan, line 233: what with comparing by paths? and not strings?

Cheshiriks: Yes. Fixed it

kgevorkyan: Just compare the paths, starting from the directory, which is the **parent** for the root `save.toml`, think this should work

save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
```diff
@@ -247,6 +247,17 @@ class FixPlugin(
         )
     }
 
+    private fun isComparingTestAndCopy(test: Path, testCopy: Path): Boolean {
+        val testCopyPath = testCopy.relativeTo(FileSystem.SYSTEM_TEMPORARY_DIRECTORY)
+            .toString()
+            .substringAfter(Path.DIRECTORY_SEPARATOR)
+            .toPath()
+
+        val testCopy = test.createRelativePathToTheRoot(testConfig.getRootConfig().directory).toPath()
```
kgevorkyan, line 256: Should it be called `testCopy`? I think it original test file

save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
```diff
@@ -247,6 +247,17 @@ class FixPlugin(
         )
     }
 
+    private fun isComparingTestAndCopy(test: Path, testCopy: Path): Boolean {
+        val testCopyPath = testCopy.relativeTo(FileSystem.SYSTEM_TEMPORARY_DIRECTORY)
```
kgevorkyan, line 251: Could you also write a comments, why do we calculate paths in this way?
Some words like: `testCopyPath` stored in `tmpDir`, holdind the whole hierarchy of original file
while `test` file comes to us with path , starting from `testRootPath`, so we compare them in such way

kgevorkyan (approved):

PR #480, opened 2023-01-24
Changed test

Cheshiriks:
Changed test


kgevorkyan:
Why do we need to change this?

Cheshiriks:
> Why do we need to change this?

This is for a test with the same names. Now the tests have the same content. Even if the names are mixed up, the test will succeed because the content is the same.

kgevorkyan:
> Even if the names are mixed up, the test will succeed **because the content is the same.**

But it was succeed with different content too, why did you said, that it is **because the content is the same.**?

kgevorkyan (approved):

PR #481, opened 2023-01-24
Update okio to 3.3.0

0x6675636b796f75676974687562:


kgevorkyan:
For now, I can't merge https://github.com/saveourtool/save-cli/pull/471
since macos is failed. probably because of that `ktoml` use  kotlin `1.8.0` too
So now I'm working on https://github.com/saveourtool/save-cli/pull/470 and I guess we don't need an separate PR for `okio`
and can unclude in `470`

kgevorkyan:
Covered in https://github.com/saveourtool/save-cli/pull/471/

PR #484, opened 2023-02-03
Add log4j-slf4j2-impl for FixPlugin

kgevorkyan:
### What's done:
* Add log4j-slf4j2-impl for FixPlugin


save-plugins/fix-plugin/build.gradle.kts
```diff
@@ -12,6 +12,9 @@ kotlin {
                 implementation(libs.multiplatform.diff)
                 implementation(libs.kotlinx.serialization.core)
                 implementation(libs.sarif.utils)
+                // log engine for sarif-utils library
+                implementation(libs.log4j.core)
```
nulls, line 16: I'm not sure that we should add implementation on this level. `fix-plugin` is library too.

but `save-cli` module is a cli application, I think we should add `libs.log4j.core` and `libs.log4j.slf4j2.impl` only there

kgevorkyan: Make sense, moved

nulls (approved):

PR #498, opened 2023-02-15
Execute tool in sarif mode before report checking

kgevorkyan:
### What's done:
* Execute tool before checking the sarif report, this logic allows:
  1) Support checking `.sarif` from stdout
  2) Rewrite existing sarif report, if `actual[Warnings|Fix]SarifFileName` was provided and something were changed in report

* Minor refactoring
* Add possible config for one of the test (however it's not required, since `save-cli` will look into stdout by default)

Part of https://github.com/saveourtool/save-cli/issues/497

### DIFF
--- a/examples/kotlin-diktat/sarif-actual/save-warnings-actual.sarif
+++ b/examples/kotlin-diktat/sarif-actual/save-warnings-actual.sarif
@@ -1,0 +1,50 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "originalUriBaseIds": {
+        "%SRCROOT%": {
+          "uri": "file://D:/projects/"
+        }
+      },
+      "results": [
+        {
+          "level": "error",
+          "locations": [
+            {
+              "physicalLocation": {
+                "artifactLocation": {
+                  "uri": "src\\kotlin\\EnumValueSnakeCaseTest.kt",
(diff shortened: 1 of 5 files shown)


examples/kotlin-diktat/sarif-actual/save-warnings-actual.sarif
```diff
@@ -0,0 +1,155 @@
+{
```
kgevorkyan, line 1: seems, that `execute warn-plugin with actual warnings from SARIF` test didn't worked at all, since there wasn't any sarif file for this test

kgevorkyan, line 1: I added this config for consistency: by default warn plugin in sarif mode will look into stdout of tool for `.sarif` report
But, it also possible to provide such report via file, like this


In aim to avoid misunderstandings, I add this config, however it's not required as been said

save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
```diff
@@ -188,6 +188,7 @@ class FixPlugin(
      * @param testCopyToExpectedFilesMap list of paths to the copy of tests files, which will be replaced by modificated files
      * @return updated list of test files copies
      */
+    // TODO: support case, when sarif report is provided via stdout
```
kgevorkyan, line 191: will cover in next PR

Cheshiriks (approved):

PR #499, opened 2023-02-16
Support getting sarif report from stdout in FixPlugin

kgevorkyan:
### What's done:
* Support getting sarif report from stdout in FixPlugin

### DIFF
--- a/save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
+++ b/save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
@@ -1,3 +1,5 @@
+@file:Suppress("FILE_IS_TOO_LONG")
+
 package com.saveourtool.save.plugins.fix
 
 import com.saveourtool.save.core.config.ActualFixFormat
@@ -229,6 +231,7 @@ class FixPlugin(
         return fixedTestCopyToExpectedFilesMap
     }
 
+    @Suppress("SAY_NO_TO_VAR")
     private fun getSarifFile(
         executionResultStdout: List<String>,
         fixPluginConfig: FixPluginConfig,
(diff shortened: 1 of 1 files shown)


save-plugins/fix-plugin/src/commonMain/kotlin/com/saveourtool/save/plugins/fix/FixPlugin.kt
```diff
@@ -219,6 +231,40 @@ class FixPlugin(
...
+    @Suppress("SAY_NO_TO_VAR")
+    private fun getSarifFile(
+        executionResultStdout: List<String>,
+        fixPluginConfig: FixPluginConfig,
+        testsPaths: List<Path>,
+        tmpDirName: Path,
+    ): Pair<Path, Boolean> {
+        var isSarifFileEmulated = false
+
+        val sarifFile = if (fixPluginConfig.actualFixSarifFileName != null) {
```
Cheshiriks, line 243: Do we even need a case when we are given actual sarif file? Could there be such a case?

kgevorkyan: of course, actual sarif file in our terms here - sarif file with fixes, which we need to apply to test files

the **expected** files in fix plugin provided via `.kt` files

kgevorkyan: I think it more rare case, when sarif report provided via stdout, rather than via file

kgevorkyan: > of course, actual sarif file in our terms here - sarif file with fixes, which we need to apply to test files
> 
> the **expected** files in fix plugin provided via `.kt` files

However, we can also support expected fixes by sarif https://github.com/saveourtool/save-cli/issues/500

Cheshiriks: Ok, agree

Cheshiriks (approved):

PR #503, opened 2023-02-22
Make `publish...` tasks depend on `sign...` tasks

### LINKED ISSUE #502: Bug in release action
hm, one more problem while creating a shanpshot @petertrr 

image

_Originally posted by @kgevorkyan in https://github.com/saveourtool/save-cli/issues/501#issuecomment-1439705340_

petertrr:
Successful run: https://github.com/saveourtool/save-cli/actions/runs/4241944409

All `publish...` tasks share Javadoc artifact (until we set up proper Javadoc generation), so this PR makes all of them dependent.

Closes #502

### DIFF
--- a/buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
+++ b/buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
@@ -14,7 +14,7 @@ import org.gradle.api.publish.maven.tasks.AbstractPublishToMaven
 import org.gradle.api.tasks.bundling.Jar
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.configure
-import org.gradle.kotlin.dsl.create
+import org.gradle.kotlin.dsl.register
 import org.gradle.kotlin.dsl.extra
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
```diff
@@ -134,6 +136,13 @@ private fun Project.configureSigning() {
         logger.lifecycle("The following publications are getting signed: ${extensions.getByType<PublishingExtension>().publications.map { it.name }}")
         sign(*extensions.getByType<PublishingExtension>().publications.toTypedArray())
     }
+
+    tasks.withType<PublishToMavenRepository>().configureEach {
```
kgevorkyan, line 140: publishToMavenLocal also have type `PublishToMavenRepository`?

kgevorkyan (approved):
👍

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
petertrr: Yes, I believe all `publish*To*Repository` are registered in a similar way.

PR #527, opened 2023-04-24
Fixed JGit usage in gradle

nulls:
### What's done:
- added search of .git when status is checked


sanyavertolet (approved):

nulls:
@kgevorkyan I guess `sarif-utils` needs similar fix. I think it relates to run gradle inside `buildSrc` where rootProject is `buildSrc` and it's not a root of git repository. It's not clear why it's not failed in scope of PR, but it fails here https://github.com/saveourtool/save-cli/actions/runs/4784311435.

But, `save-cloud` fails in PR with upgrading reckon...

kgevorkyan:
> @kgevorkyan I guess `sarif-utils` needs similar fix. I think it relates to run gradle inside `buildSrc` where rootProject is `buildSrc` and it's not a root of git repository. It's not clear why it's not failed in scope of PR, but it fails here https://github.com/saveourtool/save-cli/actions/runs/4784311435.
> 
> But, `save-cloud` fails in PR with upgrading reckon...

will take it

PR #533, opened 2023-05-25
[Draft] Introduce junit test runner

kgevorkyan:


save-common/build.gradle.kts
```diff
@@ -17,5 +17,11 @@ kotlin {
                 implementation(libs.okio.fakefilesystem)
             }
         }
+
+        val jvmTest by getting {
+            dependencies {
+                implementation("junit:junit:4.13.2")
```
kgevorkyan, line 23: https://www.baeldung.com/junit-5-runwith#migrating-from-a-junit-4-based-runner

PR #565, opened 2024-02-14
Updating Sarif and removing github okio

akuleshov7:


kgevorkyan (approved):
should be carefully updated and probably connected with https://github.com/saveourtool/save-cli/pull/556

nulls (approved):

PR #568, opened 2024-02-27
Odd file names in actual result

### LINKED ISSUE #566: Silently skips detected actual warning that related test which are not tested
The `save-cli` allows to set pattern for file name in pattern. It can lead to problem when pattern is incorrect (a full path vs file name) and save-cli silently skips such test (0 actual).
Need to add detecting that actual result provided for not existed test

nulls:
### What's done:
- added checking actual result

It closes #566


akuleshov7 (approved):
good

PR #569, opened 2024-02-27
Supported ExpectedWarningsFormat.PLAIN

### LINKED ISSUE #567: Process warnings in plain format from the file
Right now, if we have a project, we just only can process a SARIF format, we need to think of other expected formats

nulls:
### What's done:
- supported reading expected results from external plain file

It closes #567

### DIFF
--- a/save-plugins/warn-plugin/src/commonMain/kotlin/com/saveourtool/save/plugin/warn/WarnPlugin.kt
+++ b/save-plugins/warn-plugin/src/commonMain/kotlin/com/saveourtool/save/plugin/warn/WarnPlugin.kt
@@ -4,7 +4,6 @@ import com.saveourtool.save.core.config.ActualWarningsFormat
 import com.saveourtool.save.core.config.ExpectedWarningsFormat
 import com.saveourtool.save.core.config.TestConfig
 import com.saveourtool.save.core.files.createFile
-import com.saveourtool.save.core.files.findFileInAncestorDir
 import com.saveourtool.save.core.files.getWorkingDirectory
 import com.saveourtool.save.core.files.readLines
 import com.saveourtool.save.core.logging.describe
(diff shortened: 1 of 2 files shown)


save-plugins/warn-plugin/src/commonMain/kotlin/com/saveourtool/save/plugin/warn/WarnPlugin.kt
```diff
@@ -258,30 +258,52 @@ class WarnPlugin(
...
-            collectWarningsFromSarif(warnPluginConfig, originalPaths, fs, workingDirectory)
-        } catch (e: Exception) {
-            throw SarifParsingException("We failed to parse sarif. Check the your tool generation of sarif report, cause: ${e.message}", e.cause)
-        }
-        copyPaths.associate { copyPath ->
-            copyPath.name to warningsFromSarif.filter { it.fileName == copyPath.name }
+    ): WarningMap {
+        val expectedWarningsFileName: String by lazy {
+            warnPluginConfig.expectedWarningsFileName
+                ?: throw IllegalArgumentException("<expectedWarningsFileName> is not provided for expectedWarningsFormat=${warnPluginConfig.expectedWarningsFormat}")
```
kgevorkyan, line 275: now we always expect the file? may be restore to the
`if (warnPluginConfig.expectedWarningsFormat == ExpectedWarningsFormat.SARIF || ExpectedWarningsFormat.Plain `
?

kgevorkyan (approved):

save-plugins/warn-plugin/src/commonMain/kotlin/com/saveourtool/save/plugin/warn/WarnPlugin.kt
nulls: it's a lazy variable to re-use this variable for `ExpectedWarningsFormat.SARIF` and `ExpectedWarningsFormat.PLAIN`

kgevorkyan: oh, I see

